PR #3, opened 2023-02-21
Update README: add libusb package

zoobab:
add libusb otherwise minichlink would not build.


cnlohr:
Though only required on Debian.  In Windows you build the native tools.

PR #10, opened 2023-03-03
add NHC-Link042

NgoHungCuong:
NHC-Link is flash programming tool.
Firmware based on STM32F042F6P6.


cnlohr:
This PR is a very large PR for this repo.  I would love to support this programming mechanism one way or another.  That would be great, but I would rather interface to it, rather than be the host of the specific programmer. 

Do you have code to show interfacing to it via USB?

Charles

NgoHungCuong:
you can see on https://github.com/NgoHungCuong/ch32v003fun/blob/master/NHC-Link042/main.cpp
and firmware https://github.com/NgoHungCuong/ch32v003fun/blob/master/NHC-Link042/Firmware/User/NHC_WCH_SDI.c
It is easy to use:
I use STM32F042 as USB to 1-wire adapter:
- Init protocol: setup pin PA13 to output open drain ---> connect to CH32V003
Write 0xA0 to EP1
- Deinit: setup pin PA13 to input
Write 0xA1 to EP1
- Read debug register:
Write 0xA2 + register address (1 byte) to EP1
Read EP 0x81
- Write debug register:
Write 0xA3 + register address (1 byte) + data (4 bytes) to EP1
- Read memory location:
Write 0xA4 + address (32 bits) + bus width (8 bits, 16 bits, 32 bits...) to EP1
Read EP 0x81
- Write memory location:
Write 0xA5 + address (32 bits) + data (32 bits) + bus width (8 bits, 16 bits, 32 bits...) to EP1
Detail in file https://github.com/NgoHungCuong/ch32v003fun/blob/master/NHC-Link042/main.cpp
Line 105 to 206
* Firmware processing USB data:
https://github.com/NgoHungCuong/ch32v003fun/blob/master/NHC-Link042/Firmware/LibUSB/LibUSB.c
Line 358 to 405

cnlohr:
Do you use Discord, @NgoHungCuong ?  I would like to discuss this.

Today I wrote an ESP32-S2 USB-based programmer.  I would like to support an STM32F042 programmer!

After my experiments today, a few things stood out to me:

1. I am much happier with the USB device being lower-level, in that it allows for scripted read / write and delays.
2. Are you able to use your device in Windows without any drivers or admin access?  I notice you are using a bulk endpoint instead of feature requests, I've never noticed that working?  Do you have an example of how to access it?
3. Side-note: In order to make get/set requests in Windows without a driver or admin access, you can add specific feature reports to your descriptor, see this: https://github.com/cnlohr/esp32s2-cookbook/blob/master/ch32v003programmer/components/tinyusb/tinyusb/src/class/hid/hid_device.h#L339-L357
4. I am still a little uncomfortable hosting a programmer inside this repo.  I am mostly interested:
  a. In making a tool that can conveniently use any programmer, and expose a port to openocd, and/or GDB.
  b. Supporting a tiny build environment/library for the CH32V003 so you don't need to haul around a mountain of HAL.
  c. Maintaining a library that can use low-level commands ( `NHC_WchSdi_Write`/`NHC_WchSdi_Read`) to expose high-level functionality.
5. I really like one of the features of your code, that you have a sort of high-level "please do these things."  I like that it can be separated out from the underlying `NHC_WchSdi_Write`/`NHC_WchSdi_Read`.  And I think this would be wise.

What are your goals submitting the programmer to this project instead of having the programmer be its own?

NgoHungCuong:
Yes, @cnlohr
I would like to discuss this.

NgoHungCuong:
my discord id is ngohungcuong#6211

PR #20, opened 2023-03-21
Prevent overwrite of HSITRIM default-value in SystemInit48HSI()

cathiele:
When calling the `SystemInit48HSI()`-function to set the frequency to 48 MHz the` HSITRIM`-Value in `RCC->CTLR` is set to 0 and the clock is running slower than expected. 

The Datasheet states that the setting of the PLL has to be done before setting `PLLON`.

This PR fixes this problem for me and i moved the activation of the PLL one line down.

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -807,8 +807,8 @@ asm volatile(
 void SystemInit48HSI( void )
 {
 	// Values lifted from the EVT.  There is little to no documentation on what this does.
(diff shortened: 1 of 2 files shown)


cnlohr:
How do you feel about making it so it |'s in a parameter?  I'd rather not have the extra `lw` for the bootcode under the default behavior.  But I am amenable to either.

cathiele:
What do you think about the updated version? In this way i can define the HSITRIM value and if it is not provided it uses the default one.

I did not test this version until now.

cnlohr:
Put extra parens around `HISTRIM` so if there is weird math in the `#define` it'll be fine.

But yes, I like this, a lot.

PR #26, opened 2023-03-30
fix: "minichlink -u" segmentation fault

cathiele:
When running:

```
minichlink -u
``` 

i get a segmentation fault. With this fix i can use ```minichlink``` to "unbrick" my 8-pin ch32v003 successfuly


cnlohr:
thank you!

Issue #30, opened 2023-04-10
Add Windows build specific libgcc.a to all existing Makefiles so all examples compile under Windows


EmbeddedMan:

EmbeddedMan:
Pull Request #31 ready for review and merge.

PR #31, opened 2023-04-10
Added Windows specific lines to all example Makefiles 

EmbeddedMan:
For all makefiles that didn't already have the proper code. They get the right copy of libgcc.a now. All examples now compile out of the box on Windows.

Fix for issue #30


cnlohr:
I believe a much better way to do this would be to say `-L../..` why doesn't that work? / why not add that?

EmbeddedMan:
> I believe a much better way to do this would be to say `-L../..` why doesn't that work? / why not add that?

I was just following Max's suggestion - I don't know enough about command line gcc to know about this better way of doing it.

So you're saying that I can replace the block of code I put into each Makefile with a -L../..? Where would I put that change in each Makefile?

PR #35, opened 2023-04-12
Added option so Arch Linux will pull in zicsr support

wrickert:
Added '.option arch, +zicsr\n" ' so that Arch Linux will pull in support for zicsr.


recallmenot:
oh yes please, I'm on manjaro, with riscv64-unknown-elf-gcc from AUR make complains "Error: unrecognized opcode `csrw mstatus,a0', extension `zicsr' required"
no idea how to compile here, had to dual-boot ubuntu but I'd rather not.

recallmenot:
can confirm your patch works on manjaro! thank you very much!!!

PR #36, opened 2023-04-12
GPIO example and library

recallmenot:
Added a "GPIO" example and a simple arduino-like library for stupid like me.

Thank you @cnlohr for making the ch32v003 fun!!
Would never have understood it enough to make this on my own!

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -765,7 +765,8 @@ void handle_reset()
 .option norelax\n\
 	la gp, __global_pointer$\n\
 .option pop\n\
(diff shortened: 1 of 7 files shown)


cnlohr:
Why are you making the change to ch32v003fun.c?   That already exists.

cnlohr:
see this:
https://github.com/cnlohr/ch32v003fun/commit/3c4fdca3a75d8c30b2c0aaf2607f7a78f9fb93cb

recallmenot:
> Why are you making the change to ch32v003fun.c? That already exists.

I used it locally but couldn't figure out how to exclude that from the push.
I'm a noob at all this, sorry.

PR #41, opened 2023-04-13
GPIO example expanded by ADC functions, additional ADC example

recallmenot:
hope this is ok..
though I noticed that the ADC input seems to be lagging behind slightly (like a slowly-charging and discharging capacitor), maybe I made an error somewhere.

### DIFF
--- a/examples/GPIO/wiring.c
+++ b/examples/GPIO/wiring.c
@@ -1,6 +1,6 @@
 //#include <stdio.h>
 
-#include "wiring_digital.h"
+#include "wiring.h"
 #include <stdint.h>
 
 
@@ -20,7 +20,7 @@ enum GPIOports getPort (enum GPIOpins pin) {
 
 
 
-void enablePort(enum GPIOports port) {
+void portEnable(enum GPIOports port) {
 	// Enable GPIOs
 	switch (port) {
 		case port_A:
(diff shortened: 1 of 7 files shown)


dwillmore:
A/D inputs can have some interesting design constraints.  One is source impedance.  I see you're tying the wiper of a 10K pot to the A/D input and the ends to Vcc and GND.  That gives you a max of 5K of source impedance and a minimum of zero.  I'd need to look at the datasheet and see what the suggested/required source impedance is.

But, it might also be a configuration issue, I wouldn't rule that out.  Next time I have a few minutes, I'll look into the source impedance possibility if you'd like.

cnlohr:
Would you like me to accept this as-is for now?

dwillmore:
@cnlohr I assume that was addressed to @recallmenot, but if not, I don't see a reason to wait.  This repo doesn't use a release based process, so we can easily commit any fixes we may find in the future.

recallmenot:
actually "10k" was a default value I pulled out my *** (sounds reasonable), what I used in the video I think was a 1k because I had it on hand.
VCC was 3.3V from the LinkE, maybe that influenced it?

You can go ahead and merge, would be nice if you looked into it though, I'm at a loss.

In GPIO and GPIO_analogRead Both "wiring" libraries are identical so people don't have to wonder why they have an "inferior one".

Had to inject the ADC stuff into the existing library because it already has a pinMode fn and I don't want there to be two.

PR #51, opened 2023-04-23
Redo reverted patches for MacOS 

mengstr:
I have tested the Makefile on MacOS and Linux. I assume it will still work on windows since that define is not changed.


cnlohr:
Thanks for fixing this.

PR #53, opened 2023-04-26
Support more types of OLED displays

emeb:
This PR expands support of OLED display dimensions from the original 128x32 to include 128x64 and 64x32 types commonly available with the same SSD1306 controller. Display type is selectable at compile time with preprocessor macros and the demo code automatically adapts to utilize the full display area.


cnlohr:
Looks good!

PR #61, opened 2023-04-29
basic SPI library + nRF24L01+ example

recallmenot:
what I've been up to the past week.
any improvements welcome.


cnlohr:
Are you in the Discord?  If not, we should chat about this.  My general feelings are

1. In the Makefile, we now put `all : flash` first.
2. It would be better to reform the SPI as a single-file-header.  Many of the functions should be `static inline`.  If there are functions that are not appropriate to static inline, then they can be guarded with a `MINICH32VSPI_IMPLEMENTATION`
3. Please name SPI.c something different, more like `ch32v003_spi.h` 
4. The NRF24L01 should be fine being separate.

Charles

PR #66, opened 2023-05-03
SPI boss fight lvl 3

recallmenot:
* makefiles to current standard
* SPI.h + SPI.c -> ch32v003_SPI.h
   * converted runtime branches to compiletime branches via `#if defined`
   * converted runtime math to compile-time math
   * added compile-time checks for #defines
   * CH32V003_SPI_IMPLEMENTATION implementation block now ineffective
      * was able to static inline all functions
   * clear user instructions in library
   * all functions declared in easy-to-read manner where user can find them
   * clear library structure
* nrf24l01.c and nrf24l01.h further improved function naming
* writes to port pins made write-only instead of read-write

* main README.md added info on how to make without sudo

I hope this is more appropriate for this project.


cnlohr:
So a few things, sorry this is trickling so slowly:
1. Tabs, not spaces for this repo. Your commit is inconsistent.
2. I've gone back-and-forth with suggesting a refactor of things like CH32V003_SPI_NSS_SOFTWARE_PC3.  But I think what you have is acceptable.  I don't need to keep running you around.
3. 

```c
while(!(SPI1->STATR & SPI_STATR_TXE));
```

Should be 

```c
while(!(SPI1->STATR & SPI_STATR_TXE));
```

4. We should try to have common code.  Maybe a libs folder?  I will discuss.

cnlohr:
Also, do not commit minichlink binaries.

PR #70, opened 2023-05-05
SPI boss fight lvl 5.1

recallmenot:
* makefiles to current standard
* SPI.h + SPI.c -> ch32v003_SPI.h
   * converted runtime branches to compiletime branches via `#if defined`
   * converted runtime math to compile-time math
   * added compile-time checks for #defines
   * CH32V003_SPI_IMPLEMENTATION implementation block now ineffective
      * was able to static inline all functions
   * clear user instructions in library
   * all functions declared in easy-to-read manner where user can find them
   * clear library structure
   * exterminated stray spaces in preprocessor #define requirements
* nrf24l01.c and nrf24l01.h
   * further improved function naming
   * not ported to tabs because of synchronicity to original (platform agnostic) library
* writes to port pins made write-only instead of read-write
* Arduino files not ported to tabs because `.arduino15/preferences.txt` will by default nuke tabs to spaces on edit because of default `editor.tabs.expand=true`

* main README.md added info on how to make without sudo

* removed binaries minichlink and minichlink.so and added them to .gitignore because making any example, they are reliably built nowadays, compile time is ~1s on my weaksauce i5-5300U (dual-core ultrabook processor released 8 yrs ago) so it's not as annoying as compiling riscv-newlib for 60 minutes at 100% CPU use. This way they won't ever get committed again and users always have a fresh version when pulling from repo. hope testing workflows won't have an issue with this, but apparently, they don't.
* minichlink makefile reverted to original state
* added nRF library author to RX readme

I hope this is even more appropriate for this project.


cnlohr:
Good enough for now - I will update this from main.

PR #77, opened 2023-05-07
centrifuge: separated branching GPIO

recallmenot:
I also want to add a second branchless (speedy) GPIO library all static-inline but that will take time. For now, we have this.


cnlohr:
This is very confusing to see from you.  This also looks a little rough on the eyes.  We should brainstorm a cleaner way to write all this out.  One thing that's nice is the symmetry of the GPIO structures.

```c
#define GPIOA_BASE                              (APB2PERIPH_BASE + 0x0800)
#define GPIOB_BASE                              (APB2PERIPH_BASE + 0x0c00)  // Doesn't exist
#define GPIOC_BASE                              (APB2PERIPH_BASE + 0x1000)
#define GPIOD_BASE                              (APB2PERIPH_BASE + 0x1400)
```

So we could have something where your individual GPIOs could be marked off with a base address difference + a position of the bit, i.e.

PORTC5 could be `0x1005` and the code could say:

```c
#define GPIO_ON( PORTVAL ) \
  ((GPIO_TypeDef *)((PORTVAL&0xf000)+APB2PERIPH_BASE)->BSHR = 1<< (PORTVAL&0xFF);
```

cnlohr:
Let me know how you feel about this, @recallmenot

recallmenot:
You mean accessing everything through enums? :D
I originally did that so all pins could be iterated over in sequence, no matter the port, and that functions, at runtime, could determine which ones to toggle.
I get that this is not the right approach for most cases, where we know at compile time which pins we want to use how and when.
It's just the case that I wrote it in this style first, thus I separated it before going on to make the branchless version.
Ideally, I'd like to have two versions of the library, one with my (stupid) software branching and a branchless one that will compile down to just the raw instructions.
Doing it like you showed is a great idea, I feared I'd have to write an on and an off function per pin but solving it through defines like this seems like a great option, thanks!
And should the branchless version also be able to cover the cases where software decides the pins I'll gladly delete this one myself :D

PR #81, opened 2023-05-07
Add instructions for building libgcc.a and update the binary. Fixes #74.

### LINKED ISSUE #74: Sources for the included libgcc.a
Hi Charles,

thank you for working on the ch32v003fun SDK! It's very exciting to be able to use CH32V003 without any proprietary software (if combined with [aappleby/PicoRVD](https://github.com/aappleby/PicoRVD)).

Would it be possible to provide a link to the sources for the included [libgcc.a](https://github.com/cnlohr/ch32v003fun/tree/master/misc) and the way you built it? I appreciate the included binary as it's very convenient to have everything in place, but I would also like to know how to update this libgcc.a, when necessary.

krasin:
The new libgcc.a is built using the provided instructions using 7f8b34c01dd806e49e0538b2d3779f753963e774 revision of the RISC-V toolchain.


cnlohr:
Thank you, @krasin this is some good stuff.

PR #88, opened 2023-05-09
Minor build improvements

cw2:
Minor build improvements:
- #include <stdio.h> moved to ch32v003fun.h
- added build_all.cmd to build all examples

### DIFF
--- a/ch32v003fun/ch32v003fun.mk
+++ b/ch32v003fun/ch32v003fun.mk
@@ -1,5 +1,5 @@
 
-PREFIX:=riscv64-unknown-elf
+PREFIX?=riscv64-unknown-elf
 
 CH32V003FUN?=../../ch32v003fun
 MINICHLINK?=../../minichlink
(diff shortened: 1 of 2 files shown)


cnlohr:
@cw2 I am deliberately adding `#include <stdio.h>` so that people who look at the code are consciously aware they can just `printf(...)` from the code.  Like there is no technical reason to add it, but I wanted to express it from a intellectual standpoint.

Is there a technical reason to remove it?

cw2:
I see, good point - no need to remove `#include <stdio.h>` then (I have removed/reverted the commit).

Also, I have split `.option arch, +zicsr`, will do a separate PR associated with the respective issue.

cnlohr:
Blehhhhhhhh I can't get this to compile on windows with `del /Q /F` 

What Make are you using?

cw2:
I do not have any extra tools installed except what comes with the toolchain - there is make.exe in the bin directory...

cnlohr:
I must be using a different make.  They appear to be mutually exclusive.  I spent about an hour trying to get something to autodetct if I even could execute `del` but I couldn't figure it out.  Maybe I need to do it the other way around, i.e. see if I can execute rm.

Well, don't install anything extra.  I want to figure this out.

PR #91, opened 2023-05-11
Add support for math using softfp library.

ghent360:
Some math functions from libm complain about __errno() - added one.
Add magic incantation to the linked libs so it does not complain about missing symbols (yes list -lgcc twice)
Add new OPTFLAGS variable so user Makefiles can overwrite the optimization settings - useful for debugging

Note: removed /usr/include/newlib include path - I venture to guess most users would have compilers installed somewhere else, who knows whats in that folder and how compatible it is. Things seem to compile fine w/o it.

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -14,6 +14,10 @@
 
 int errno;
 
+int * __errno(void) {
+    return &errno;
+}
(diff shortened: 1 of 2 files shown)


ghent360:
Looks like the default riscv compiler is missing the math lib. Will try another time.

PR #96, opened 2023-05-14
Fix restarting vscode debugger

AlpyneDreams:
Moved the flash command to deploySteps which makes it runs on restart.


cnlohr:
This is great.  Thanks.

PR #97, opened 2023-05-14
Added zicsr arch option. Closes #89

### LINKED ISSUE #89: Error 'unrecognized opcode csrrs' when compiling self_modify_code example with gcc 12.x
Unfortunately, due to a breaking change in RISC-V ISA 20191213 (https://groups.google.com/a/groups.riscv.org/g/sw-dev/c/aE1ZeHHCYf4)

>... the csr read/write (csrr*/csrw*) instructions and fence.i instruction has separated from the `I` extension, become two standalone extensions: Zicsr and Zifencei;...

gcc 12.x now requires `-march=rv32ec_zicsr` in order to compile `csrr*` instructions. Although it is an easy fix when supporting just the latest gcc 12.x risc-v toolchains, the question is how to do it while keeping backward compatibility with older gcc toolchains? E.g. a conditional parameter assignment in makefile based on gcc version?

cw2:


cnlohr:
Thank you!

PR #104, opened 2023-05-20
make fields in (gpio) registers available via structs

kabel42:
this is a first attempt to do this for the gpio registers

are you generally interested in something like this?
do the names make sense?
To me this is easier to read than bitshifts but the enums are very verbose

structs are autogenerated here https://github.com/kabel42/ch32v003

### DIFF
--- a/ch32v003fun/ch32v003fun.h
+++ b/ch32v003fun/ch32v003fun.h
@@ -196,15 +196,127 @@ typedef struct
 } OB_TypeDef;
 
 /* General Purpose I/O */
+typedef enum
+{
+    GPIO_CFGLR_MODE_IN,
+    GPIO_CFGLR_MODE_OUT_10MHz,
+    GPIO_CFGLR_MODE_OUT_2MHz,
+    GPIO_CFGLR_MODE_OUT_50MHz
+} GPIO_MODE_TypeDef;
+typedef enum
+{
+    GPIO_CFGLR_CNF_IN_ANALOG = 0,
+    GPIO_CFGLR_CNF_IN_FLOAT = 1,
+    GPIO_CFGLR_CNF_IN_PULL_UP_DOWN = 2,
+    GPIO_CFGLR_CNF_OUT_PP = 0,
+    GPIO_CFGLR_CNF_OUT_OD = 1,
+    GPIO_CFGLR_CNF_OUT_AF_PP = 2,
+    GPIO_CFGLR_CNF_OUT_AF_OD = 3,
+} GPIO_CNF_TypeDef;
 typedef struct
 {
-    __IO uint32_t CFGLR;
+    __IO union {
+      uint32_t CFGLR;
+      struct CFGLR_t {
+        GPIO_MODE_TypeDef MODE0 :2;
+        GPIO_CNF_TypeDef CNF0 :2;
+        GPIO_MODE_TypeDef MODE1 :2;
+        GPIO_CNF_TypeDef CNF1 :2;
+        GPIO_MODE_TypeDef MODE2 :2;
+        GPIO_CNF_TypeDef CNF2 :2;
+        GPIO_MODE_TypeDef MODE3 :2;
+        GPIO_CNF_TypeDef CNF3 :2;
+        GPIO_MODE_TypeDef MODE4 :2;
+        GPIO_CNF_TypeDef CNF4 :2;
+        GPIO_MODE_TypeDef MODE5 :2;
+        GPIO_CNF_TypeDef CNF5 :2;
+        GPIO_MODE_TypeDef MODE6 :2;
+        GPIO_CNF_TypeDef CNF6 :2;
+        GPIO_MODE_TypeDef MODE7 :2;
+        GPIO_CNF_TypeDef CNF7 :2;
+      } CFGLR_bits;
+    };
     __IO uint32_t CFGHR;
-    __IO uint32_t INDR;
-    __IO uint32_t OUTDR;
-    __IO uint32_t BSHR;
-    __IO uint32_t BCR;
-    __IO uint32_t LCKR;
+    __IO union {
+      uint32_t INDR;
+      struct INDR_t {
+        uint32_t IDR0 :1;
+        uint32_t IDR1 :1;
+        uint32_t IDR2 :1;
+        uint32_t IDR3 :1;
+        uint32_t IDR4 :1;
+        uint32_t IDR5 :1;
+        uint32_t IDR6 :1;
+        uint32_t IDR7 :1;
+        uint32_t :24;
+      } INDR_bits;
+    };
+    __IO union {
+      uint32_t OUTDR;
+      struct OUTDR_t {
+        uint32_t ODR0 :1;
+        uint32_t ODR1 :1;
+        uint32_t ODR2 :1;
+        uint32_t ODR3 :1;
+        uint32_t ODR4 :1;
+        uint32_t ODR5 :1;
+        uint32_t ODR6 :1;
+        uint32_t ODR7 :1;
+        uint32_t :24;
+      } OUTDR_bits;
+    };
+    __IO union {
+      uint32_t BSHR;
+      struct BSHR_t {
+        uint32_t BS0 :1;
+        uint32_t BS1 :1;
+        uint32_t BS2 :1;
+        uint32_t BS3 :1;
+        uint32_t BS4 :1;
+        uint32_t BS5 :1;
+        uint32_t BS6 :1;
+        uint32_t BS7 :1;
+        uint32_t :8;
+        uint32_t BR0 :1;
+        uint32_t BR1 :1;
+        uint32_t BR2 :1;
+        uint32_t BR3 :1;
+        uint32_t BR4 :1;
+        uint32_t BR5 :1;
+        uint32_t BR6 :1;
+        uint32_t BR7 :1;
+        uint32_t :8;
+      } BSHR_bits;
+    };
+    __IO union {
+      uint32_t BCR;
+      struct BCR_t {
+        uint32_t BR0 :1;
+        uint32_t BR1 :1;
+        uint32_t BR2 :1;
+        uint32_t BR3 :1;
+        uint32_t BR4 :1;
+        uint32_t BR5 :1;
+        uint32_t BR6 :1;
+        uint32_t BR7 :1;
+        uint32_t :24;
+      } BCR_bits;
+    };
+    __IO union {
+      uint32_t LCKR;
+      struct LCKR_t {
+        uint32_t LCK0 :1;
+        uint32_t LCK1 :1;
+        uint32_t LCK2 :1;
+        uint32_t LCK3 :1;
+        uint32_t LCK4 :1;
+        uint32_t LCK5 :1;
+        uint32_t LCK6 :1;
+        uint32_t LCK7 :1;
+        uint32_t LCKK :1;
+        uint32_t :23;
+      } LCK_bits;
+    };
 } GPIO_TypeDef;
 
 /* Alternate Function I/O */
--- a/ch32v003fun/ch32v003fun.h
+++ b/ch32v003fun/ch32v003fun.h
@@ -227,7 +227,7 @@ typedef union {
 		GPIO_CFGLR_PIN_MODE_Typedef PIN6 :4;
 		GPIO_CFGLR_PIN_MODE_Typedef PIN7 :4;
 	};
-} CFGLR_t;
+} GPIO_CFGLR_t;
 typedef union {
 	uint32_t __FULL;
 	const struct {
@@ -241,7 +241,7 @@ typedef union {
 		uint32_t IDR7 :1;
 		uint32_t :24;
 	};
-} INDR_t;
+} GPIO_INDR_t;
 typedef union {
 	uint32_t __FULL;
 	struct {
@@ -255,7 +255,7 @@ typedef union {
 		uint32_t ODR7 :1;
 		uint32_t :24;
 	};
-} OUTDR_t;
+} GPIO_OUTDR_t;
 typedef union {
 	uint32_t __FULL;
 	struct {
@@ -278,7 +278,7 @@ typedef union {
 		uint32_t BR7 :1;
 		uint32_t :8;
 	};
-} BSHR_t;
+} GPIO_BSHR_t;
 typedef union {
 	uint32_t __FULL;
 	struct {
@@ -292,7 +292,7 @@ typedef union {
 		uint32_t BR7 :1;
 		uint32_t :24;
 	};
-} BCR_t;
+} GPIO_BCR_t;
 typedef union {
 	uint32_t __FULL;
 	struct {
@@ -307,7 +307,7 @@ typedef union {
 		uint32_t LCKK :1;
 		uint32_t :23;
 	};
-} LCKR_t;
+} GPIO_LCKR_t;
 typedef struct
 {
 	__IO uint32_t CFGLR;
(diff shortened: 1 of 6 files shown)


dwillmore:
I like the look of this.  Does it compile down to sane code that only sets the specified fields?  

You used implicit numbering in GPIO_MODE_TypeDef but not in the other structs.

It's good that you point out the implicit read-modify-write and how to do multiple things at once.  Withtout that, people could semantically think that these are inexpensive operations and not bundle them up--leading to bloated code.  Since these registers are volatile, the compiler isn't allowed to clean up after peoples mistakes.

I'm curious if @cnlohr has feedback on this.

kabel42:
I've tried to recreate the direct_gpio example with structs.
With Os it seems similar, but it's hard to say as it is very scrambled.
With Og it's mostly the same code but it seems to struggle sometimes:

```
		// Use upper bits of BSHR to RESET output
		GPIOC->BSHR = (1<<(16+1)); // RESET GPIO C1
 172:	000207b7          	lui	a5,0x20
 176:	c81c                	sw	a5,16(s0)
```
vs
```
		// Use upper bits of BSHR to RESET output
		GPIOC->BSHR_bits = (const struct BSHR_t) {
			.BR1 = 1,
		};
 13c:	00042823          	sw	zero,16(s0) # 40011010 <__global_pointer$+0x20010810>
 140:	481c                	lw	a5,16(s0)
 142:	00020737          	lui	a4,0x20
 146:	8fd9                	or	a5,a5,a4
 148:	c81c                	sw	a5,16(s0)
```

as far as i understand this, it now tries to store 32 bits, where before it only stored the upper 16 bit?

dwillmore:
That's the kind of thing I'm worried about.

dwillmore:
Ahh, it doesn't know about how the register reads as zero and you can only write ones to it.  It's tring to read it and add in the bit you wanted which takes a lot longer.  (I don't read RISC-V very well, yet)  So I may be wrong.

kabel42:
maybe we should skip the write only registers and only implement structs for config registers, where gcc generates exactly the same code:
```
		// Modify the OUTDR register directly to SET output
		GPIOC->OUTDR |= 1<<(4);    // SET GPIO C4
 160:	445c                	lw	a5,12(s0)
 162:	0107e793          	ori	a5,a5,16
 166:	c45c                	sw	a5,12(s0)
```
vs
```
		// Modify the OUTDR register directly to SET output
		GPIOC->OUTDR_bits.ODR4 = 1;
 13e:	445c                	lw	a5,12(s0)
 140:	0107e793          	ori	a5,a5,16
 144:	c45c                	sw	a5,12(s0)
```

dwillmore:
That seems like a much safer proposal.

kabel42:
I can get it to generate the same code, but now the c code is no longer more readable
```
		GPIOC->BSHR = 1<<(1);      // SET GPIO C1
 154:	40011437          	lui	s0,0x40011
 158:	4789                	li	a5,2
 15a:	c81c                	sw	a5,16(s0)
```
``` 
		const struct BSHR_t BS1 = {
			.BS1 = 1,
		};
		GPIOC->BSHR_bits = BS1;
 124:	40011437          	lui	s0,0x40011
 128:	4789                	li	a5,2
 12a:	c81c                	sw	a5,16(s0)
```

dwillmore:
Strange that it does it right with a variable, but not with a direct assignment.

dwillmore:
I summon the power of @duk-37

duk-37:
Can't test anything right now but I suspect this may have to do with the compiler being too smart for its own good wrt. uninitialized data

By any chance are there unnamed bitfields, e.g.:
```c
struct foo {
   int :4;
   int b: 4;
};
```

in this struct? I know there are in some others from the same header. If so, could you try giving them names and testing again?

kabel42:
i have changed it to
``` c
struct BSHR_t {
        uint32_t BS0 :1;
        uint32_t BS1 :1;
        uint32_t BS2 :1;
        uint32_t BS3 :1;
        uint32_t BS4 :1;
        uint32_t BS5 :1;
        uint32_t BS6 :1;
        uint32_t BS7 :1;
        uint32_t RESERVED1:8;
        uint32_t BR0 :1;
        uint32_t BR1 :1;
        uint32_t BR2 :1;
        uint32_t BR3 :1;
        uint32_t BR4 :1;
        uint32_t BR5 :1;
        uint32_t BR6 :1;
        uint32_t BR7 :1;
        uint32_t RESERVERD2:8;
      } BSHR_bits;
```
but the compiler still wants to write zero to the register and then read it back before doing anything else.
Maybe volatile is somehow rubbing of on the data?

jnk0le:
to do write only by this bitfield you need to eg.  do the "shadow register" as TI does [11]

Also, type punning by union bitfields is very fragile. It's UB in C++ and implementation specified or something like that in C and the compilers can perform narrower accesses to a 32bit only peripherals; par 7 in [11],[12],[13].

▪ [11] http://staff.ii.pw.edu.pl/kowalski/dsp/F28x/F2808_page/spraa85a.pdf
▪ [12] https://stackoverflow.com/questions/67340350/bitfield-write-size
▪ [13] https://stackoverflow.com/questions/42171429/force-gcc-to-access-structs-with-words

jnk0le:
ok now I see that the issue is with
```
GPIOC->BSHR_bits = (const struct BSHR_t) {
	.BR1 = 1,
};
```
try writing to uint32 union member instead of bitfield one

kabel42:
If i understand the TI appnote shadow registers work like https://github.com/cnlohr/ch32v003fun/pull/104#issuecomment-1555973181
I don't want to force that on the user. If that is the best we can do, i'd prefer to write set/clear functions/macros.
Maybe ST style: 
``` c
static inline void setPin(GPIO_TypeDef* gpio, enum pin pin)
{ gpio->BSHR = pin}
setPin(GPIOA, GPIO_pin1|GPIO_pin2);
```
or
``` c
static inline void setPin(GPIO_TypeDef* gpio, int pin)
{ gpio->BSHR = 1<<pin}
setPin(GPIOA, 1);
```

type punning should work in c++ with `extern "C"`

unnamed union, i want to preserve backward compatibility, e.g. `GPIOA->BSHR = 1<<4;` should still work.

kabel42:
@jnk0le i'd like to do that, but i'd have to name the union and that would break old code. Or i'm missing something obvious. :)

jnk0le:
writing more than one bit at a time seems again efficient

```
int main()
{
    GPIOC->BSHR_bits = (const struct BSHR_t) {
        .BR0 = 1,
        .BR1 = 1,
    };
}

main:
        li      a4,196608
        li      a5,1073811456
        sw      a4,16(a5)
        li      a0,0
        ret
```

and here is the broken write: https://godbolt.org/z/naYbczrvs

duk-37:
> writing more than one bit at a time seems again efficient
> 
> ```
> int main()
> {
>     GPIOC->BSHR_bits = (const struct BSHR_t) {
>         .BR0 = 1,
>         .BR1 = 1,
>     };
> }
> 
> main:
>         li      a4,196608
>         li      a5,1073811456
>         sw      a4,16(a5)
>         li      a0,0
>         ret
> ```
> 
> and here is the broken write: https://godbolt.org/z/naYbczrvs

Right, okay, back on computer, this is GCC being terrible and turning single-bit volatile writes into a write-to-zero-then-bitwise-or, which is _technically_ legal (though I have no idea why you'd do it..) because the C standard is evil and the only real guarantee w/ volatile is that it's an "observable side-effect." meaning it can't be reordered relative to other side effects but that everything else is fair game. *sigh*

duk-37:
Anyway, here's one of the jankiest pieces of code I've ever written.

```c
volatile struct BSHR_t BS;

// https://gcc.gnu.org/pipermail/gcc/2020-November/234216.html
#define GCC_IS_EVIL_VOLATILE_WRITE(field, value) { const typeof(((void)0, field)) __v = value; field = __v; }

int main() {
    GCC_IS_EVIL_VOLATILE_WRITE(BS, {
        .BR1 = 1
    })
}
```

duk-37:
A somewhat more reasonable fix is to actually write the u32 directly: https://godbolt.org/z/5s558rM7b

```c
union BSHR_t {
    struct {
        uint32_t BS0 :1;
        uint32_t BS1 :1;
        uint32_t BS2 :1;
        uint32_t BS3 :1;
        uint32_t BS4 :1;
        uint32_t BS5 :1;
        uint32_t BS6 :1;
        uint32_t BS7 :1;
        uint32_t RESERVED1:8;
        uint32_t BR0 :1;
        uint32_t BR1 :1;
        uint32_t BR2 :1;
        uint32_t BR3 :1;
        uint32_t BR4 :1;
        uint32_t BR5 :1;
        uint32_t BR6 :1;
        uint32_t BR7 :1;
        uint32_t RESERVERD2:8;
    };
    uint32_t full;
};

volatile uint32_t BS;

int main() {
    BS = (union BSHR_t) {
        .BR1 = 1
    }.full;
}
```

kabel42:
gcc doesn't like
``` c
    __IO union BSHR_t {
      uint32_t BSHR;
      struct {
        uint32_t BS0 :1;
        uint32_t BS1 :1;
        uint32_t BS2 :1;
        uint32_t BS3 :1;
        uint32_t BS4 :1;
        uint32_t BS5 :1;
        uint32_t BS6 :1;
        uint32_t BS7 :1;
        uint32_t :8;
        uint32_t BR0 :1;
        uint32_t BR1 :1;
        uint32_t BR2 :1;
        uint32_t BR3 :1;
        uint32_t BR4 :1;
        uint32_t BR5 :1;
        uint32_t BR6 :1;
        uint32_t BR7 :1;
        uint32_t :8;
      } BSHR_bits;
    };
```
```
struct_direct_gpio.c:53:22: error: 'GPIO_TypeDef' has no member named 'BSHR'
   53 |                 GPIOC->BSHR = 1;
      |                      ^~
```

duk-37:
Functional example here -- https://godbolt.org/z/39rxxTsx1

Hopefully this solves various issues with the current approach -- the main one being that embedding the bits into the GPIO struct directly is really error prone as we've seen so far.

Essentially requires moving all the structs to outside the big GPIO structs, turning them into unions, keeping the existing values as u32s, then having a nice define for everything. That gives us a useful abstraction without losing finer-grained control or (hopefully) confusing the compiler.

kabel42:
i think we lost the write byte example, for future reference, this is how it goes wrong for structs and equivalent with bitshift:
```
	GPIOD->OUTDR_bits.BAIT = 0xff;
 632:	57fd                	li	a5,-1
 634:	40f406a3          	sb	a5,1037(s0)
	GPIOD->OUTDR = 0xff<<16;
 638:	00ff07b7          	lui	a5,0xff0
 63c:	c4dc                	sw	a5,12(s1)
```

kabel42:
Test: What happens to the register if it gets a sb?
C:
``` c
printf("CFGLR: %lX\n", GPIOC->CFGLR); // -> CFGLR: 44444444
GPIOC->CFGLR_bits.BAIT2 = 0x00;
printf("CFGLR: %lX\n", GPIOC->CFGLR);
```
ASM/LST:
```
	GPIOC->CFGLR_bits.BAIT2 = 0x00;
 624:	000400a3          	sb	zero,1(s0) # 40011001 <_eusrstack+0x20010801>
```
Output:
```
CFGLR: 44444444
CFGLR: 0
```

... = 0xFF:
```
CFGLR: 44444444
CFGLR: FFFFFFFF
```

cnlohr:
A few thoughts.

1. Tabs, not spaces on this repo. (Please update, when you get a chance)
2. I think it would be kinda fun/cool to include even the write-only.
3. Don't you need to add `__attribute__((packed))`?
4. A lot of people really like this way of addressing values.  I want to add it.
5. What version of GCC are you using, @kabel42 ?

I really would like to merge this as soon as these are addressed.  It's something I think would be great!

kabel42:
1: no problem, would it be possible to have a .clang-format for that?
2: what do you think about these functions, to keep the compiler from adding extra bloat?
``` c
static inline void GPIOset(GPIO_TypeDef *gpio, uint8_t pins) {gpio->BSHR = pins;}
static inline void GPIOreset(GPIO_TypeDef *gpio, uint8_t pins) {gpio->BCR = pins;}
static inline void GPIOsetReset(GPIO_TypeDef *gpio, uint8_t set_pins, uint8_t reset_pins) {gpio->BSHR = set_pins|(reset_pins<<16);}
```
3: bitflields are packed, afaik you only need that to not align stuff like uint8
5: I'm using community/riscv64-elf-gcc 12.2.0-1 from the arch repos and `export PREFIX=riscv64-elf `

I'll tidy this up and add another commit

cnlohr:
1. Bleh... I should probably get one together at some point.  I don't know how to make them or where they go.  Do you know much about them?  for now, can just spaces->tabs.  For the most part that's the only real style I feel that strongly about, just because I don't use an IDE.  The spaces don't insert themselves.
2. I don't understand why you would have them.  Writing `GPIOC->BSHR = _____` shouldn't add bloat?
3. It should be fine, but the attribute makes it clear to onlookers.
5. Gotcha - should be ok then.

cnlohr:
Actually 3: Attribute Packed.  How do others feel?  If no one else wants it, I don't want it either.

kabel42:
2: It shouldn't, but gcc does:
```
		GPIOD->BSHR_bits = (struct BSHR_t) {.BS0 = 1};
 65e:	40011437          	lui	s0,0x40011
 662:	40040793          	addi	a5,s0,1024 # 40011400 <_eusrstack+0x20010c00>
 666:	0007a823          	sw	zero,16(a5)
 66a:	4b98                	lw	a4,16(a5)
 66c:	00176713          	ori	a4,a4,1
 670:	cb98                	sw	a4,16(a5)
```

kabel42:
staring at CFGLR i really want to change it to the following. It's not what the RM says, but it's more intuitive to me. What do you think?
``` c
typedef enum
{
	GPIO_CFGLR_IN_ANALOG = 0,
	GPIO_CFGLR_IN_FLOAT = 4,
	GPIO_CFGLR_IN_PUPD = 8,
	GPIO_CFGLR_OUT_10Mhz_PP = 1,
	GPIO_CFGLR_OUT_2Mhz_PP = 2,
	GPIO_CFGLR_OUT_50Mhz_PP = 3,
	GPIO_CFGLR_OUT_10Mhz_OD = 5,
	GPIO_CFGLR_OUT_2Mhz_OD = 6,
	GPIO_CFGLR_OUT_50Mhz_OD = 7,
	GPIO_CFGLR_OUT_10Mhz_AF_PP = 9,
	GPIO_CFGLR_OUT_2Mhz_AF_PP = 10,
	GPIO_CFGLR_OUT_50Mhz_AF_PP = 11,
	GPIO_CFGLR_OUT_10Mhz_AF_OD = 13,
	GPIO_CFGLR_OUT_2Mhz_AF_OD = 14,
	GPIO_CFGLR_OUT_50Mhz_AF_OD = 15,
} GPIO_CFGLR_PIN_MODE_Typedef;
typedef struct
{
	__IO union {
	uint32_t CFGLR;
	struct {
		GPIO_CFGLR_PIN_MODE_Typedef PIN0 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN1 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN2 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN3 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN4 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN5 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN6 :4;
		GPIO_CFGLR_PIN_MODE_Typedef PIN7 :4;
	  } CFGLR_PINS;
// ...

**Comment by duk-37** [comment]
> 1: no problem, would it be possible to have a .clang-format for that?
> 2: what do you think about these functions, to keep the compiler from adding extra bloat?
> ``` c
> static inline void GPIOset(GPIO_TypeDef *gpio, uint8_t pins) {gpio->BSHR = pins;}
> static inline void GPIOreset(GPIO_TypeDef *gpio, uint8_t pins) {gpio->BCR = pins;}
> static inline void GPIOsetReset(GPIO_TypeDef *gpio, uint8_t set_pins, uint8_t reset_pins) {gpio->BSHR = set_pins|(reset_pins<<16);}
> ```
> 3: bitflields are packed, afaik you only need that to not align stuff like uint8
> 5: I'm using community/riscv64-elf-gcc 12.2.0-1 from the arch repos and `export PREFIX=riscv64-elf `
> 
> I'll tidy this up and add another commit

2. at at point just use a macro
3. no, you have a uint32 in your union field. also your bitfields are aligned to u32 because that's the type specified on the left. again I strongly suggest to ditch the union completely; see the last example i sent.

**Comment by kabel42** [comment]
I'm happy to go either way, my reasoning is:
2: Why risk the headache of debugging macros when a static inline works
3: I don't think it's very readable and vscode autocomplete can't handle it
might as well add more defines and have something like
``` c
GPIOD->CFGLR = GPIO_CFGLR_MODE0_OUT_10MHz | GPIO_CFGLR_CNF0_OUT_PP;
GPIOD->BSHR = GPIO_Pin_0 | GPIO_Pin_1;
```

cnlohr:
Ugh this gives me so much anxiety.  I have seen some people really preferring this style of using registers, so I want to support it, but, I don't have the deep expertise to know the right avenue to do it to avoid weird leaky abstraction issues.  If you guys can continue to work through it, I'd really appreciate it.

AlexanderMandera:
This PR currently does not compile in a C++ context.
I have found a fix, see in https://github.com/cnlohr/ch32v003fun/pull/105#issuecomment-1557772402.

Please keep this in mind as you move forward to ensure that the C++ CI action doesn't fail, thank you.

jnk0le:
>We could also go the easy way and just add some defines, but we would miss out on the type checking
>```
>GPIOD->CFGLR = GPIO_CFGLR_MODE0_OUT_10MHz | GPIO_CFGLR_CNF0_OUT_PP;
>GPIOD->BSHR = GPIO_Pin_0 | GPIO_Pin_1
>```

`BSHR` aka `BSRR` should use the `GPIO_BSRR_BS_0` and `GPIO_BSRR_BR_0` like defs. Otherwise you need a magic 16bit shift to do do gpio clearing.

The most missing definitions are the `*_Pos` ones like in more recent ST headers for f0 and newer peripherals.
With only those we get to something like that:
```
// assume reset state of MODER4 field
GPIOA->MODER |= (0b01 << GPIO_MODER_MODER4_Pos);

// assume unknown state MODER4 field
// GPIO_MODER_MODER4_Msk is the same thing as GPIO_MODER_MODER4
// just less confusing
GPIOA->MODER = (GPIOA->MODER & GPIO_MODER_MODER4_Msk) | (0b01 << GPIO_MODER_MODER4_Pos);
```

there are also things like systick which doesn't have any bit definitions, so you have to initialize it with magic value.

kabel42:
I'm out of good ideas and i have no idea how to do this with macros so here is an attempt with lots of static inline function.
@duk-37, if you can do this with macros i'd happily that, but i can't do it.

This hopefully also fixes the c++ problems.

duk-37:
> I'm out of good ideas and i have no idea how to do this with macros so here is an attempt with lots of static inline function. @duk-37, if you can do this with macros i'd happily take that, but i can't do it.
> 
> The examples compile to sane asm that should be as fast as the bitshift approach.
> 
> This hopefully also fixes the c++ problems.

`__VA_ARGS__` to the rescue. :)
```c
#define DYN_GPIO_READ(field) ((const GPIO_##field##_t) { .__FULL = GPIO.field })
#define DYN_GPIO_WRITE(field, ...) GPIO.field = ((const GPIO_##field##_t) __VA_ARGS__).__FULL

uint32_t Test;
int main() {
    Test = DYN_GPIO_READ(BSHR).BR0;
    DYN_GPIO_WRITE(BSHR, { .BR1 = 1, .BR2 = 1 });
}```

**Comment by kabel42** [comment]
for single values, maybe something like this?
``` c
#define something(gpio, field, reg, val) \
{const GPIO_##field##_t tmp = DYN_GPIO_READ(gpio, field); \
tmp.##reg = val; DYN_GPIO_WRITE(gpio, field);}
```

duk-37:
> 

Yep, more thinking about the multi-value case. But honestly should be fine for now tbh

cnlohr:
Just making sure - everyone is green lighting this current PR, right?

kabel42:
normal struct can only modify one value at a time, so this should be fine

i'll add an example of doing this by hand for multiple values

i'd like to add gpio to the macro so you can specify the port like
``` c
    DYN_GPIO_WRITE(GPIOC, BSHR, { .BR1 = 1, .BR2 = 1 });
```

kabel42:
> Just making sure - everyone is green lighting this current PR, right?

let me push the macro version first unless someone prefers 33 lines of static inline functions just for gpio

kabel42:
asm looks sane, direct_gpio and structured_direct_gpio generate the same code

i'm fine with merging this

PR #105, opened 2023-05-21
Fix error when using C++ virtual methods

AlexanderMandera:
When using ch32v003fun with C++, there is an error when using virtual methods,
which is shown in the example program "cpp_virtual_methods" in the examples folder.

Without the usage of `__libc_init_array`, the virtual table of the class instance `Example` is not initialized correctly:

```
(gdb) info vtbl Example
vtable for 'ExampleClass' @ 0x0 (subobject @ 0x20000004):
[0]: 0x11e0006f
[1]: 0x0 <InterruptVectorDefault()>
```

In this case, a call to a virtual method will result to an invalid call and the program won't work as expected.

Here, the example program will only print out `Begin example`,
but does not print any values which are implemented in overridden virtual methods.
It seems like the MCU resets here.

In other environments like with my Arduino Core, the hard fault interrupt handler was called
and the program gets into an infinite loop.

With the added macro `CPLUSPLUS`, an implemention auf `__libc_init_array` is called in `handle_reset()`.

After that change, the virtual table is initialized correctly and new values are printed as expected:

```
(gdb) info vtbl Example
vtable for 'ExampleClass' @ 0x212c (subobject @ 0x20000004):
[0]: 0x1fb0 <ExampleClass::doNewLine()>
[1]: 0x1fda <ExampleClass::printValue(int)>
```

This also reflects the startup code for C++ inside the `OpenWCH/ch32v003` repository:
https://github.com/openwch/ch32v003/blob/main/C%2B%2B/Use%20MRS%20Create%20C%2B%2B%20project-example/CH32V003F4P6/Startup/startup_ch32v00x.S#L160

Another small change I did, was to add `TARGET_EXT` in `ch32v003fun.mk`, so my target file can be a C++ file and it is not limited to C.

The wrong behavior with vtables can be tested by removing the `CPLUSPLUS` macro in the example program's Makefile.
If another macro name other than `CPLUSPLUS` or even another solution is better in this case, then let me know :)

P.S.: I have tested my changes in my Arduino Core with my HardwareSerial implementation
and it seems to work fine there. So when this PR is merged, I can use upstream ch32v003fun as a submodule there. 👍

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .pio
 .vscode
+.idea
 *.elf
 *.hex
 *.lst
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -660,6 +660,10 @@ mini_pprintf(int (*puts)(char*s, int len, void* buf), void* buf, const char *fmt
 	Copyright 2023 Charles Lohr, under the MIT-x11 or NewBSD licenses, you choose.
 */
 
+#ifdef CPLUSPLUS
+// Method to call the C++ constructors
+void __libc_init_array(void);
+#endif
 
 int main() __attribute__((used));
 void SystemInit( void ) __attribute__((used));
@@ -801,6 +805,12 @@ asm volatile(
 	addi a1, a1, 4\n\
 	bne a1, a2, 1b\n\
 2:\n" );
+
+#ifdef CPLUSPLUS
+    asm volatile(
+        "call %0 \n\t"        // Call __libc_init_array function
+        : : "i" (__libc_init_array) : );
+#endif
 
 	SETUP_SYSTICK_HCLK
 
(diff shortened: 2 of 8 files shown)


cw2:
> If another macro name other than `CPLUSPLUS` or even another solution is better in this case, then let me know :)

There is a standard `__cplusplus` macro/symbol defined when C++ compiler is in use.

AlexanderMandera:
> > If another macro name other than `CPLUSPLUS` or even another solution is better in this case, then let me know :)
> 
> There is a standard `__cplusplus` macro/symbol defined when C++ compiler is in use.

I already tried this out and to my knowledge that doesn't really work,
because this macro is only defined if it is used in a C++ file directly.

So when I include the header file inside a C++ file, the `__cplusplus` macro is set.
But because the C file is compiled seperately, the macro is not defined by the compiler.

That is why I have used another macro that is both used in C and C++ files.

cnlohr:
I am not sure where my comments I'm definitely OK with keeping it a separate macro!  And just #defining it in the build system.

Also, tabs, not spaces in this repo.

AlexanderMandera:
Sure, I have changed the indentation in my IDE now.
I also made the `__libc_init_array` inline with the previous assembly like you mentioned in Discord.

cnlohr:
Sounds good! I will merge.

cnlohr:
@AlexanderMandera it looks like this is not compiling with CI - do you mind trying to address it?

https://github.com/cnlohr/ch32v003fun/actions/runs/5044741252/jobs/9058473174?pr=104

AlexanderMandera:
@cnlohr According to this [StackOverflow post](https://stackoverflow.com/questions/17637392/anonymous-union-can-only-have-non-static-data-members-gcc-c), you are not allowed to define nested types inside an anonymous union.

What I tried, was to remove the names of the nested structs like `LCKR_t`

from
```c
__IO union {
	uint32_t LCKR;
	struct LCKR_t {
```
to
```c
__IO union {
	uint32_t LCKR;
	struct {
```

Then the example and the other C-only programs compile just fine.
I looks like that this only applies when including `ch32v003fun.h` in a C++ context.

PR #106, opened 2023-05-22
Use CMake to compile `minichlink` with macOS

crosstyan:
The [origin makefile](https://github.com/cnlohr/ch32v003fun/blob/ab3581a2dac6a23574cede351c550aab77d1982f/minichlink/Makefile#L19-L25) won't compile on my M1 Mac. Though add compilation condition to [linux/in.h](https://github.com/cnlohr/ch32v003fun/blob/ab3581a2dac6a23574cede351c550aab77d1982f/minichlink/microgdbstub.h#L75) could just make it compile, I still think CMake would be a better build toolchain. 

Adding other OS support would be trivial but I don't have other environment for now.


cnlohr:
Please make any changes to the Makefile you need to make guarded in something like the following:

```make
    UNAME_S := $(shell uname -s)
    ifeq ($(UNAME_S),Linux)
        CCFLAGS += -D LINUX
    endif
    ifeq ($(UNAME_S),Darwin)
        CCFLAGS += -D OSX
    endif
```

Having the parallel build system will confuse people and requiring people to install CMake to build this is not acceptable.

cnlohr:
(Though I really do want to have OSX support!)

cnlohr:
I really would like OSX support for this, though!

crosstyan:
The path you provide is correct in my environment, though I would prefer to use `pkg-config` since hard coding path should be avoided.

```makefile
	ifeq ($(OS_NAME),darwin)
		LDFLAGS:=-lpthread -lusb-1.0 -framework CoreFoundation -framework IOKit
		CFLAGS:=-O0 -Wall -Wno-asm-operand-widths -Wno-deprecated-declarations -Wno-deprecated-non-prototype -D__MACOSX__
		INCLUDES:=$(shell pkg-config --cflags-only-I libusb-1.0)
		LIBINCLUDES:=$(shell pkg-config --libs-only-L libusb-1.0)
		INCS:=$(INCLUDES) $(LIBINCLUDES)
	endif
```

PR #107, opened 2023-05-22
SSD1306 added drawImage function

recallmenot:
I've been working on (image) compression and the drawImage function for the @emeb SSD1306 was a necessary side product.
As the display operates in vertical mode with us looking at it horizontally, still every pixel needs to be set.
I managed a tiny speed increase by eliminating the call to drawPixel but the main benefit of the function is to use black or white as transparency and the ability to draw ontop of the buffer contents using bitmath, think sprites!

Of course the SSD1306_LOG_IMAGE can be stripped away.

I created a separate example as the SSD1306 library belongs to @emeb and I didn't want to bulldoze it.
This way you can modify the original to your own liking and later delete this example; it's the most respectful way of going about this I could think of.


emeb:
Why don't you just add that function to the original i2c_oled example instead of duplicating the whole thing?

cnlohr:
@recallmenot 

1. Are you up for unifying the files and then moving the header files to the `extralibs` folder and referencing them from this project that way? "Bulldoze" implies major changes, but it should be possible to add with only minor changes.
2. Is there an issue with using the `make gdbserver`?

recallmenot:
Ok, thought it wasn't welcome, will do that right now.
I won't even try moving it to extralibs myself, might do something stupid.
Btw, what's the `make gdbserver` thing about? Hearing about it here the first time :D
3 days ago I had my first encounter with GDB

PR #110, opened 2023-05-22
SSD1306 drawImage

recallmenot:
I've been working on (image) compression and the drawImage function for the @emeb SSD1306 was a necessary side product.
As the display operates in vertical mode with us looking at it horizontally, still every pixel needs to be set.
I managed a tiny speed increase by eliminating the call to drawPixel but the main benefit of the function is to use black or white as transparency and the ability to draw ontop of the buffer contents using bitmath, think sprites!

Of course the SSD1306_LOG_IMAGE can be stripped away.


cnlohr:
It looks good on my end! What do you think @emeb ?

emeb:
Looks good to me!

PR #112, opened 2023-05-23
converted all example include paths

recallmenot:


cnlohr:
Thanks for going back and fixing this up!

PR #113, opened 2023-05-23
provide include paths to clangd

recallmenot:


cnlohr:
What benefits does this offer over `Make`?  Additional parallel build systems confuse people and require more churn when making modifications.  I.e. if someone adds something, they will only add it to their current build system and other build systems will break.

PR #114, opened 2023-05-23
added support for clangd #2

recallmenot:

___
moved the build_scripts


recallmenot:
sorry forgot the makefile modification

PR #115, opened 2023-05-23
added support for clangd #3

recallmenot:

___
moved the build_scripts

___
this time with clangd in the makefile, like suggested


cnlohr:
Looks good except you deleted the bin.  I would like to keep the bin for people who "Just want to flash a binary" to make sure it works.

image

recallmenot:
ah so that was there by intent!
I thought it was a leftover

PR #117, opened 2023-05-23
Update wiring.c

prosper00:
fixed off-by-1 error addressing pins on port A: the enum starts with pin_A1 (i.e. 'pin_A1 = 0'), so, attempts to address this pin need to be incremented by 1.

Could also update the enum definition to include a pin_A0, but this pin doesn't actually exist on any current packages.

### DIFF
--- a/examples/GPIO/wiring.c
+++ b/examples/GPIO/wiring.c
@@ -45,7 +45,7 @@ void pinMode(enum GPIOpins pin, enum GPIOpinMode mode) {
 	
 	if (pin <= pin_A2) {
 		GPIOx = GPIOA;
-		PinOffset *= pin;
+		PinOffset *= (pin + 1);
 	}
 	else if (pin <= pin_C7) {
 		GPIOx = GPIOC;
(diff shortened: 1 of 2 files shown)


recallmenot:
nice catch!
though I already made a non-branching one over on [ch32v003fun_wildwest](https://github.com/recallmenot/ch32v003fun_wildwest/tree/main)/tree/main
wasn't halal for this repo though

cnlohr:
Hmm, I didn't notice this the first time, but also, it should probably not be solved this way anyway.  It should probably be solved with shifts. 


```c
void pinMode(enum GPIOpins pin, enum GPIOpinMode mode) {
	GPIO_TypeDef * GPIOx;
	uint16_t PinOffset;

	if (pin <= pin_A2) {
		GPIOx = GPIOA;
		PinOffset = (pin + 1)<<2;
	}
	else if (pin <= pin_C7) {
		GPIOx = GPIOC;
		PinOffset = (pin - 2)<<2;
	}
	else if (pin <= pin_D7) {
		GPIOx = GPIOD;
		PinOffset = (pin - 10)<<2;
	}
	else {
		return;
	}
```

Even though the compiler *should* do this on its own, it may not. 

Total side-question:  Should we be padding the enum so that it would contain the unconnected A* and B* pins, so the code could just simply be:

```c
void pinMode(enum GPIOpins pin, enum GPIOpinMode mode) {
	GPIO_TypeDef * GPIOx;
	uint16_t PinOffset;

	GPIOx = GPIOA+(pin>>3);
	PinOffset = (pin & 0x7)<<2;
```

?

prosper00:
Oh yeah, makes sense. I didn't even notice that there weee multiplies in pinmode(). I'll update the PR tomorrow

PR #118, opened 2023-05-24
GPIO example fixes: fixed port A off-by-one, replaced multiplies with bitshifts, Padded GPIO enum def's

prosper00:
@cnlohr - is this more like what you'd had in mind for the wiring.* fixes?


cnlohr:
Are you ok with this @recallmenot ?

cnlohr:
Also, I think it would be pretty cool to include this in `extralibs`.

recallmenot:
> rt of - I realized that it may be easier to just use fixed points.
> 
> i.e.

that looks much better
do so if you want!

BobRyan530:
Just a note that the reference manual makes no mention of GPIOB, but there does seem to be a gap in the register addressing for it. I don't know if it makes more sense to keep it in for padding purposes or remove it for consistency with the documentation
`0x40011Cxx`

dwillmore:
Other related chips do have GPIOB and there is hope to add support for them
to this library.

prosper00:
yes, I'll update the PR later today

cnlohr:
Wonderful!

maxgerhardt:
The code in this PR has broken digitalWrite() and pinMode() on anything else but GPIOA. The expression to compute the value of the GPIO register is plain wrong.

https://github.com/cnlohr/ch32v003fun/blob/421c0c489eadb0b9de5328781dc72f79a7a5c7fb/examples/GPIO_analogRead/wiring.c#L42-L46

This code assumes that "GPIOA + 0" = GPIOA (fine), "GPIOA + 1 = GPIOB" (wrong), "GPIOA + 2 = GPIOC". Since GPIO is of type `GPIO_TypeDef*`, aka a pointer to the struct definition of a GPIO peripheral, the sizeof which is 8 bytes, the expression "GPIO + 1" would just be advanding 8 bytes after GPIOA to get what it thinks is GPIOB, and +16 to get what it thinks is GPIOC. 

**This is wrong**. The peripheral address are
```
GPIOA 0x40010800 
GPIOC 0x40011000
```
so if you have an imaginatory GPIOB, the distance between each peripheral address is 0x400 bytes. Not 8.

One way to fix the expression is to say
```c
	GPIOx = ((void*)GPIOA+(pin>>3)*0x400);
```
but I think the multiplication will need way more computational time than just doing a `switch(pin >> 4)` and returning GPIOA, GPIOC, etc. as needed.

As is, this PR has broken the example.

PR #119, opened 2023-05-24
added support for clangd #4

recallmenot:

___
moved the build_scripts

___
this time with clangd in the makefile, like suggested

___

had an adventure with git restoring the blink.bin

in case anyone ever needs to perform similar surgery on a branch
```
travel back one commit, keeping file changes intact
do as often as needed
git reset --soft HEAD~

remove changed file from staging (pending for commit)
git reset -- <path>

restore file to state before current commit
git restore <path>

create new divergent head
git commit --amend -m "message"

push (unsafe (only ever do on own branch!!!!))
git push --force
```
always check with `git status` between each step!!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 *.map
 minichlink/minichlink
 minichlink/minichlink.so
+compile_commands.json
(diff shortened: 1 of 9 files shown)


cnlohr:
Excellent!

PR #122, opened 2023-05-24
time macros 3

recallmenot:
* added macros to get current time according to systick
 * clarified SysTick mode description
 * moved all the sections to do with time together so they're in one spot
 * fixed typos
 * moved #ifndef __ASSEMBLER__ back where it was (moved by misstake)

### DIFF
--- a/ch32v003fun/ch32v003fun.h
+++ b/ch32v003fun/ch32v003fun.h
@@ -4958,9 +4958,17 @@ static inline uint32_t __get_SP(void)
 extern "C" {
 #endif
 
-// You can use SYSTICK_USE_HCLK, if you do, you will have a high-resolution
-// however it will limit your max delay to 44 seconds before it will wrap
-// around.  You must also call SETUP_SYSTICK_HCLK.
+/* SYSTICK info
+ * time on the ch32v003 is kept by the SysTick counter (32bt)
+ * by default, it will operate at (SYSTEM_CORE_CLOCK / 8) = 6MHz
+ * to unlock 48MHz:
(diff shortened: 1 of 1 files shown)


cnlohr:
I'm really sorry, but I am 90% sure this doesn't work the way you think it would/want it to.  Systick times have to remain native for time stamps to work right.  In addition, it is very bad for users to unwittingly call `__divdi3` in many cases as it's exceptionally slow.

An additional issue is if you divide early instead of late, it will bring about unexpected-by-use rollover issues, i.e. every 715 or 89 seconds depending on run mode.

The expectation is that users should operate in the systick time base, and there are no rollover or other unexpected issues. It's important for users to not be "surprised" by things when they don't, at runtime do what they expect.  It would be good to educate users that that is the "mode" they should operate in, and if they do, the underlying compiled firmware works out much more cleanly and simply, with less edges for people to cut themselves on.

I think it may be really helpful to improve the examples we currently have with timing.

I think chatting about these sorts of things in advance would be a really good move, to make sure that the direction of things you are doing (you are quite the tour de force) are directed effectively.

cnlohr:
To clarify for any others who get here who are not in the discord...

The appropriate way to use systicks are:

1. To fire an interrupt in the future:   `SysTick->CMP = SysTick->CNT + (microseconds * DELAY_US_TIME)`
2. To compare with a certain amount of time after an event: `if( ((int32_t)(SysTick->CNT - target_time)) > 0 ) { event is fired }`
3. To get an elapsed time: `elapsed = (uint32_t)(SysTick->CNT - last_event_time)`

PR #124, opened 2023-05-25
Input capture

kabel42:
Input capture example,
implements and uses RCC and TIM structs
renamed #define HSITRIM to HSITRIM_DEFAULT due to name collision with register RCC->CTLR->HSITRIM


cnlohr:
Did you really intend to make the changes you have here to ch32v003fun?  Please review.  I think I'm ok with them. But I don't fully understand them.

cnlohr:
Also are you on the Discord, so we can discuss?

PR #126, opened 2023-05-26
Add I2C slave library & example

renzenicolai:
This library and example show how to use the I2C peripheral in slave mode.

The library uses a one-byte address, allowing for up to 256 registers to be defined.

The first byte written to the device within a transaction determines the offset for following reads and writes, emulating a simple EEPROM.

The example will turn on a LED connected to PD0 when the LSB of register 0 is set to 1 and off when it's set to 0.

### DIFF
--- a/examples/i2c_slave/Makefile
+++ b/examples/i2c_slave/Makefile
@@ -1,0 +1,9 @@
+all : flash
+
+TARGET:=i2c_slave_example
+
+include ../../ch32v003fun/ch32v003fun.mk
+
+flash : cv_flash
+clean : cv_clean
+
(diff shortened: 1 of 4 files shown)


cw2:
> The library uses a one-byte address, allowing for up to 256 registers to be defined.

_Nitpick: The I2C address is only 7-bit (or extended 10-bit)._

renzenicolai:
This library acts like a memory chip with up to 256 registers, the address can be set freely as parameter of the setup function.

Only 7 bit addresses are supported at the moment

cnlohr:
Oops - merging went bad - this doesn't actually pass CI - you can see the CI errors.  Mind trying to get that going?

PR #129, opened 2023-05-27
Fix calculation of GPIO peripheral in wiring.c

### LINKED ISSUE #128: wiring.c GPIO referencing broken
Per https://github.com/cnlohr/ch32v003fun/pull/118#issuecomment-1565673989.

maxgerhardt:
Uses same codeblock as already used in digitalWrite, digitalRead..

Fixes #128


cnlohr:
OMG I totally missed that.

This is still wrong (though it produces the correct output)

I believe it should be 

```c
GPIOx = GPIOA+(pin>>3)*0x0400;
```

Can you test it?

cnlohr:
Resolved in #131

PR #138, opened 2023-05-29
timer2 pwm example

recallmenot:
finally it works!!
I conceptually knew timer2 gets clock from APB1, but this line cost me HOURS to write: `	RCC->APB1PCENR |= RCC_APB1Periph_TIM2;`


cnlohr:
You would not be the first to have missed that.  We may want to have some guide for stuff like that.

Thank you for persisting.

PR #141, opened 2023-05-29
Add I2C slave library and example

renzenicolai:
Should pass CI this time, apparently the main C file and output files should be named the same as the directory containing the example.


cnlohr:
ty

PR #144, opened 2023-05-31
add zicsr support to __enable_irq and __disable_irq

recallmenot:
hope this is correct,
otherwise won't compile on arch-based linux
same fix as before (not me though)


cnlohr:
Thanks

PR #147, opened 2023-06-02
Adding example remapping TIM2's output

prosper00:
Any interest in adding this example to your repo, demonstrating the use of the AFIO feature?


cnlohr:
Please switch to using printf over debug pins.  It works with all programmers now.

As far as wanting this examples, YES! I would love it.

prosper00:
Done

recallmenot:
Nice! was thinking about adding it, but figured showing where to find it would be good enough.

recallmenot:
but maybe rename it to "tim2_pwm_remap" so the names stay together in the examples list (coherence)

prosper00:
done.

recallmenot:
how did you do that without having to do a new PR?? what wizardry is this?

prosper00:
I made the changes in my fork that I initiated the PR from, then 'git add -A; git commit; git push'

recallmenot:
Nothing else? And here I was always doing a new PR for each modification on my fork.

prosper00:
> Nothing else? And here I was always doing a new PR for each modification on my fork.

No, nothing else. Git seems to be on top of it, doing its magic. The git push automatically updates the PR, triggers the CI, etc.

Issue #146, opened 2023-06-02
Prebuilt MacOS minichlink missing


mengstr:
https://github.com/cnlohr/ch32v003fun/wiki/Installation states under MacOS ```You can use the pre-compiled minichlink or go to minichlink dir and make it.```

But there is no prebuilt executable for MacOS in the repo.  Should there be one? It's kinda practical to have ready-made executables, but I guess most people will create them without any issues.

maxgerhardt:
I think it was there at some point but got deleted by a commit. But I can just add CI to build minichlink for every possible configuration and then link it.

maxgerhardt:
Can I ask you to download and try the right MacOS from https://github.com/Community-PIO-CH32V/ch32v003fun/actions/runs/5157203799 as test? 

Btw are .tar.gz or .zip files more popular for MacOs Downloads? You'd need to first unzip then untar (`tar xfv <file.tar.gz>`) the archive, then `./minichlink -h` as execution test.

mengstr:
No joy. I tried the MacOS 12 download (I'm actually on MacOS 13) and I got a zip that contained a tar.gz which ended up as a `minichlink: Mach-O 64-bit executable x86_64`.

After the usual complaints about unsigned code it ran after I approved it. But resulted in a missing library.

```
mats@Matss-MacBook-Pro Downloads % ./minichlink
dyld[69320]: Library not loaded: /usr/local/opt/libusb/lib/libusb-1.0.0.dylib
  Referenced from: <4EABE23C-602B-39A1-9247-81A79DB348EC> /Users/mats/Downloads/minichlink
  Reason: tried: '/usr/local/opt/libusb/lib/libusb-1.0.0.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/opt/libusb/lib/libusb-1.0.0.dylib' (no such file), '/usr/local/opt/libusb/lib/libusb-1.0.0.dylib' (no such file), '/usr/local/lib/libusb-1.0.0.dylib' (no such file), '/usr/lib/libusb-1.0.0.dylib' (no such file, not in dyld cache)
zsh: abort      ./minichlink
```
Which is reasonable since I'm having a M1 CPU and it runs X86 code emulated in Rosetta. But it then can't use my installed M1 libusb. (At least I think it's like that ^__^)

When I compile it myself using the Makefile I get a M1/ARM executable `./minichlink:  Mach-O 64-bit executable arm64`that loads the M1 libusb

[...]

PR #148, opened 2023-06-02
Build minichlink in CI

maxgerhardt:
Uses Github Action with a `ubuntu-latest`, `macos-12` and `macos-11` runner to produce Linux x64, Windows x64 and MacOs (Intel) x64 binaries.

Marked as draft until #146 confirms that binary is runable on MacOS.

~~Would then also re-add the `minichlink` file precompiled for MacOS.~~ It's better to not add precompiled files because we don't know whether they'll be using ARM M1 Mac or Intel Mac, so they better build it themselves.

Wiki page at https://github.com/cnlohr/ch32v003fun/wiki/Installation was updated in regards to the outdated Mac section with a strikethrough.

As always, **squash upon merge**.


cnlohr:
Is it ready to go?  I regret that I don't speak CI, so I just have to trust yall on this.  I notice it is marked as a draft.

maxgerhardt:
Ready to go.

PR #150, opened 2023-06-04
interrupt suspension functions

recallmenot:
добрые друзья,

I'd like to propose some basic interrupt suspension functions.
The intention is to temporarily disable all interrupts except the one you're interested in so it can run un-impeded and with as little delay as possible, without the need for nesting.
I hope this will be useful and is light enough to be part of ch32v003fun.

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -765,6 +765,51 @@ void InterruptVectorDefault()
 	.option   pop;\n");
 }
 
+
+
+/*
+SUSPEND ALL INTERRUPTS EXCEPT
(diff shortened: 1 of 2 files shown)


cnlohr:
Makes sense.

Have you tested this significantly?

recallmenot:
Not what I'd call extensively, though I have noticed an improvement in timing jitter over elevating IRQ priority (afaik) and it's currently the best I can do.

cnlohr:
I care more about correctness and convenience.  So if you think it's correct, I'll accept it.

recallmenot:
to the best of my knowledge, it should be.
It's not like I have hundreds of usecases to test it either...
it worked for me, I can only hope it works for everyone else too.

cnlohr:
Excellent!

PR #153, opened 2023-06-05
Add MAX7219 SPI display driver library + example and small SPI library fixes

alexhorner:


cnlohr:
@alexhorner could you rebase this off of current master?  And add an appropriate funconfig.h

Also, can you move your MAX7219 specific "extralib" into the max7219 folder?

After that, I believe I am up for accepting the PR.

cnlohr:
After much deliberation, I have decided to decline this PR.

I would rather not include drivers for other specific devices like this moving forward.  Please see https://github.com/recallmenot/ch32v003fun_wildwest and https://github.com/hexeguitar/ch32v003fun_libs for avenues forward here.

Sorry about this.

PR #155, opened 2023-06-06
GPIO branchless

recallmenot:
I'd like to share the fruits of the last few days.
This GPIO library, while hard to read (and harder to write) is branchless and heavily (ab)uses the preprocessor to output register manipulations identical to what you'd do by hand, only now with the ease of using a library.
Compiled size is comparable to hand written register manipulation: original blink is 500b, blink using this library boils down to 504 bytes (no pre-combined manipulations).
It has digital IO, ADC and PWM capabilities for both timers.
Both examples have been unified, 3 more have been added to demonstrate the full capabilities of the library.
[KITT](https://www.youtube.com/watch?v=oNyXYPhnUIs) has also been implemented.
I hope this is acceptable.


recallmenot:
solved.
now for the devastating criticism. :D

cnlohr:
As long as it's thoroughly tested, I'm pretty good ... also, please just check your overall, squashed history.  This still removes  blink.bin.  To be clear, fixing that and your saying it's good is all I need to accept!!

Bonus:
Have you compared the flash usage before/after?  I would expect this to be much better.  But iterating while monitoring those values help generate optimal solutions.

recallmenot:
Again the blink.bin bit me! I didn't delete it from my repo, I only recompiled it (brand new shiny gcc13.1) and now git thinks it needs to be wiped. I'll try re-adding it, gimmeasec.

recallmenot:
Ah yes blink.bin is back, but the version my gcc spat out is a bit smaller.
Is that bad?
Should I replace with old downloaded version?

recallmenot:
In case you can recommend any other advanced C tricks (just got my brain around nested ternary, macro-like functions, multi-line macros and macro token concatenation) , especially for abusing the preprocessor, please let me know!
This is all very new to me and I'd like to continue to learn!

recallmenot:
So now I've tested sizes.  
the old GPIO example was 640 bytes, now is 580  
the old GPIO_analogRead example was 5212 bytes! Suspect: floats.  
Then I backported my new "obfuscated" bitshift + subtraction math and it went down to 884 bytes, that's >4300 bytes for floating point math!!  
Is that really how big those functions are if you don't have an FPU?  

For reference:
```
uint8_t leds_to_turn_on = (uint8_t)(((float)(analogRead(Ain6_D6)) / 1024.f) * 8.f * 1.2 - 1.f);
```
became
```
uint16_t analog_result = GPIO_analogRead(GPIO_Ain6_D6);
analog_result = analog_result > 10 ?
		(analog_result < ((1 << 10) - 42) ? analog_result + 42 : (1 << 10))
	: 0;
uint8_t leds_to_turn_on = analog_result >> 7;
```
The GPIO_analogRead demo using the new branchless library is 836 bytes.

PR #156, opened 2023-06-08
clangd issue -32001 invalid ast

recallmenot:
This one has been bugging me for a while, clangd (language server of LLVM) had been broken by this.
I finally figured out that the generated compile_commands.json, which contains march and mabi from the gcc command, causes clangd to error out on `clangd issue -32001 invalid ast`
The solution is to generate a `.clangd` file when running `make clangd` that instructs clangd to ignore the march and mabi lines as described here:
https://github.com/clangd/clangd/issues/1582
The error is gone and clangd now works as expected.
Please confirm the .clangd file is correctly generated on WSL (should be).


GiorgosXou:
@cnlohr This is missing from `ch32fun.mk` causing the same issue. Idk where it got lost but the momment I've added those few commits\lines to my local makefile the issue was solved.

PR #162, opened 2023-06-10
Add an Arduino-based SWIO programmer

BlueSyncLine:
I implemented a quick and dirty SWIO interface based on an Arduino Uno R3 board.
For the specific boards that I have, target reset logic seems to be required, so it was added as well.

Regards,
- Mark

### DIFF
--- a/minichlink/Makefile
+++ b/minichlink/Makefile
@@ -1,7 +1,7 @@
 TOOLS:=minichlink minichlink.so
 
 CFLAGS:=-O0 -g3 -Wall
-C_S:=minichlink.c pgm-wch-linke.c pgm-esp32s2-ch32xx.c nhc-link042.c minichgdb.c
+C_S:=minichlink.c pgm-wch-linke.c pgm-esp32s2-ch32xx.c nhc-link042.c ardulink.c minichgdb.c
 
 # General Note: To use with GDB, gdb-multiarch
 # gdb-multilib {file}
--- a/minichlink/Makefile
+++ b/minichlink/Makefile
@@ -1,7 +1,7 @@
 TOOLS:=minichlink minichlink.so
 
 CFLAGS:=-O0 -g3 -Wall
-C_S:=minichlink.c pgm-wch-linke.c pgm-esp32s2-ch32xx.c nhc-link042.c ardulink.c minichgdb.c
+C_S:=minichlink.c pgm-wch-linke.c pgm-esp32s2-ch32xx.c nhc-link042.c minichgdb.c
 
 # General Note: To use with GDB, gdb-multiarch
 # gdb-multilib {file}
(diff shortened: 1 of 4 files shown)


cnlohr:
I am very interested in adding support for the most programmers possible into minichlink, however, I am hoping to link them to the actual programmer project elsewhere instead of including the programmers here. I, early on, allowed for the NCHlink042, but I would rather avoid that moving forward, since fixes would have to go through me and I'm just not an expert.

Is it OK if the firmware lives in a separate repository and link it from your driver for minichlink?

BlueSyncLine:
Certainly, I already have it in a separate repo:
https://gitlab.com/BlueSyncLine/arduino-ch32v003-swio.

You can exclude commit 0c404e2 which pertains solely to the actual
implementation and not the driver.

Regards,
- Mark

cnlohr:
> You can exclude commit [0c404e2](https://github.com/cnlohr/ch32v003fun/commit/0c404e23d862f83c3daf4ae7dc8c64a7822886e7) which pertains solely to the actual

You will need to exclude it, to use the online workflow.

Also, it not supporting Windows is a pretty black eye, that would be good.

maxgerhardt:
I'd love to see the serial code being generalized and usable for Windows too. I've worked a tiny bit on [seriallib.cpp](https://github.com/imabot2/serialib/blob/master/lib/serialib.cpp) in the past, that's basically a template on how to compile-time switch in Windows or Unix serial functions. If I remember right, `avrdude` also has a tiny serial abstraction layer written in C.

Another idea would be to generalize the firmware running on the Arduino Uno to use the general Arduino framework (and even there you can do fast port manipulation in Arduino-like way, see [here](https://github.com/arduino/ArduinoCore-avr/blob/eabd762a1edcf076877b7bec28b7f99099141473/cores/arduino/Arduino.h#L178-L184)). That way people could use many more boards as their programmer.

cnlohr:
Any chance you would update this to remove the arduino code? And add clarifications on how to build/test?  Others can then more easily test and may add Windows support.

BlueSyncLine:
@cnlohr  I removed the unnecessary commit. What clarifications on building would be desired?

AlexanderMandera:
Currently, it does not compile on Windows:

```bash
ardulink.c:5:10: fatal error: termios.h: No such file or directory
    5 | #include <termios.h>
      |          ^~~~~~~~~~~
```

You may check if the `WINDOWS` macro is defined and prevent ardulink from being included until there is proper Windows support implemented.

BlueSyncLine:
Hello,
I just spotted an issue.

The changes I committed might break the other programmers, as
MCF.TargetReset is NULL for everything except for the Ardulink...
https://github.com/cnlohr/ch32v003fun/blob/03f8a99101bcfbc23df753e3880db678958c2599/minichlink/minichlink.c#L1669

cnlohr:
already fixe dit.

BlueSyncLine:
Just saw it, apologies.
Thank you!

PR #163, opened 2023-06-10
Fix SPI Direction

gnxlxnxx:
Value was shifted 4 bits to the right


cnlohr:
Thank you for catching this.  Crazy this came from the upstream.

PR #169, opened 2023-06-13
Create XW extension encoder

duk-37:
Gives us the ability to use some of the fancy proprietary compressed instructions.

I've tested it a bunch against `objdump -M xw` and things _seem_ to line up, but there may be bugs so I'm marking this as a draft for now.

Based off of work from https://github.com/cnlohr/rv003usb/issues/11 and [here](https://github.com/CaiB/CH32V003-Architecture-Exploration/blob/main/Analysis/NewInstructionNotes.md).


cnlohr:
Any reason not to accept this?

duk-37:
nope, tested it a bit more and it should be okay

PR #170, opened 2023-06-13
Add example for writing to User Data0/1 in Option Bytes

mengstr:


recallmenot:
already did a while back but Charles said it was a project:
https://github.com/recallmenot/ch32v003fun_wildwest/tree/main/proj/flash_storage_optionbytes
https://github.com/recallmenot/ch32v003fun_wildwest/tree/main/proj/flash_storage_main

cnlohr:
And this only writes at boot.  I do like this.

Will approve.

PR #179, opened 2023-06-17
cleanup and simplify AH fix for ch32v307 read memory

bmx:
+ remove round up in -r mode to read the requested amount of bytes
+ remove the 0x307 check and stash the old ReadBinaryBlob method


alexhorner:
You appear to have completely removed the register dump code?

I left it in there with an #if 0 after discussing with cnlohr as it could end up being useful to someone

PR #180, opened 2023-06-17
Make Ardulink compilable on Windows

maxgerhardt:
Introduces a thin abstraction layer in `serial_dev.h` using a `serial_dev_t` structure.

Compile-time switches in **native** Windows or Posix (Linux, Mac) functions while keeping the return value meanings unchanged.

Succesfully compiles for Windows now, test is outstanding.


cnlohr:
caaaan you test?

maxgerhardt:
Absolutely, doing some improvements and tests right now, just opened "prematurely" to raise awareness that I'm working on this.

maxgerhardt:
grafik

Works!

maxgerhardt:
Woah, what's going in here? The `TargetReset` function pointer is set in ardulink.c, but `minichlink.c` never uses it? As such, the init() method will activate reset (1) but noone ever pulls the chip out of reset. I don't imagine that that is how it worked previoiusly.

maxgerhardt:
Okay, so https://github.com/cnlohr/ch32v003fun/commit/785ce78340cba2966fe34093f1c899c178a83cef removed the call to pull reset down to 0 again and thus broke ardulink.

How should this be handled @cnlohr? Should the init function not touch the reset line, or should the call to reset (0) be added back in, provided the reset function exist?

cnlohr:
> 

This is fine.

cnlohr:
Uuhh hmm... there are conflicts...  Any chance you could pull again?  I realized I had another tree with other adjacent changes.

maxgerhardt:
1. I've resolved the merge conflicts
2. I've re-added the reset code again with a null check so that Ardulink works when RESET is connected. Should not affect other programmers.

maxgerhardt:
Since it was heavily bugging me that the Ardulink COM port wasn't selectible via a command argument switch, I've now added in this functionality. Please carefully inspect https://github.com/cnlohr/ch32v003fun/pull/180/commits/fbab6c28e763dfd0aeb0765133ffc4ebc88d4fc9. 

It adds the concept of initialization hints for the init functions, the first member of which is the serial port. That could be in the future be expanded with more like VID/PIDs to search for, an explicit USB path to use if multiple programmers are connected to a computer, etc.

This does **break** the binary compatibility in the ABI because it adds one more parameter to the init function. If this is unwanted, I can revert.

cnlohr:
@maxgerhardt What is the purpose of TargetReset?  Everything should be 100% done via the SWIO pin, unless I'm missing something?

maxgerhardt:
I only restored what was there before (in a non-crashing way), @BlueSyncLine designed it with the reset line.

I also experienced it worked without connecting `NRST`, but if it's connected, it must be controlled, because the Arduino Uno firmware by defaults sets its pin as output and drives it low, holding the processor in reset.

PR #188, opened 2023-06-21
GPIO port operations, single-parameter pin specifiers

recallmenot:
To resolve "issues" [174](https://github.com/cnlohr/ch32v003fun/issues/174) and [172](https://github.com/cnlohr/ch32v003fun/issues/172) I'd like to present an even more **_disgusting_** GPIO library.

The GPIO port operations are a definite improvement, if not for writing to the port, then for applying the pin cfg to an entire port, much faster and slimmer than the old for loop iterating over all pins and doing a load-store each time.

For the single-parameter pin specifiers you requested:
First I tried variadic redirections due to unknowable ammount of parameters:
```
#define GPIO_pinMode(...)				GPIO_pinMode_(__VA_ARGS__)
#define GPIO_digitalWrite_hi(...)			GPIO_digitalWrite_hi_(__VA_ARGS__)
#define GPIO_digitalWrite_lo(...)			GPIO_digitalWrite_lo_(__VA_ARGS__)
#define GPIO_digitalWrite(...)				GPIO_digitalWrite_(__VA_ARGS__)
#define GPIO_digitalWrite_low(...)			GPIO_digitalWrite_low_(__VA_ARGS__)
#define GPIO_digitalWrite_0(...)			GPIO_digitalWrite_0_(__VA_ARGS__)
#define GPIO_digitalWrite_high(...)			GPIO_digitalWrite_high_(__VA_ARGS__)
#define GPIO_digitalWrite_1(...)			GPIO_digitalWrite_1_(__VA_ARGS__)
#define GPIO_digitalWrite_branching(...)		GPIO_digitalWrite_branching(__VA_ARGS__)
#define GPIO_digitalRead(...)			 	GPIO_digitalRead_(__VA_ARGS__)
```
Which worked fine in both cases (GPIO_port+n and GPIO_pin), but nuked LS parameter auto-suggestions.
That's why the new functions all have a `P` suffix and aren't variadic.
I think this is the better solution.

Though I must admit, I don't really see the point in single-parameter pin specifiers but here they are, no overhead, working fine.
Making them the only option is not a step I'm willing to take since that would mean iterating across pins would become painful.

I hope this is acceptable.

### DIFF
--- a/examples/GPIO/GPIO.c
+++ b/examples/GPIO/GPIO.c
@@ -1,13 +1,14 @@
-// 2023-06-07 recallmenot
+// 2023-06-21 recallmenot
 
 #define DEMO_GPIO_blink					1
+#define DEMO_GPIO_blink_port				0
 #define DEMO_GPIO_out					0
 #define DEMO_GPIO_in_btn				0
 #define DEMO_ADC_bragraph				0
 #define DEMO_PWM_dayrider				0
 
-#if ((DEMO_GPIO_blink + DEMO_GPIO_out + DEMO_GPIO_in_btn + DEMO_ADC_bragraph + DEMO_PWM_dayrider) > 1 \
-  || (DEMO_GPIO_blink + DEMO_GPIO_out + DEMO_GPIO_in_btn + DEMO_ADC_bragraph + DEMO_PWM_dayrider) < 1)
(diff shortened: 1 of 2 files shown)


cnlohr:
I think this is an improvement over what is currently there, still thinking through an optimal solution.

PR #194, opened 2023-06-26
GPIO_branchless 2.0: STstyle pin references

recallmenot:
From the example of the [STM32 GPIO functions](https://github.com/cnlohr/stm32f042_fun/blob/master/stm32f042cXt6_demo/firmware/systems.h#L33-L99) @cnlohr provided, this version of the GPIO library brings ST-Style pin adressing to the ch32v003, where a single byte specifies the port with its upper 4 bits, and the pin with its lower 4 bits.  
Note that both are used as `uint4_t`, not as 4 individual bits, arranging for addressing up to 16 pins on up to 16 ports.  

Here are the compiled demo sizes before and after, identical!!!
I've re-tested all demos, they work fine.  

demo                 | branchless 1.5    | branchless 2.0 
-----------------|-------------------|--------------------
GPIO_blink       | 500               | 500
GPIO_blink_port  | 448               | 448
GPIO_out         | 540               | 540
GPIO_in_btn      | 512               | 512
ADC_bargraph     | 796               | 796
PWM_dayrider     | 1148              | 1148

Even though the preprocessor output is disgusting, gcc13 seems to chew through it just fine and still optimizes the address math away entirely.  

I also integrated the separate define `GPIO_SPEED_IN` as a member of `GPIOSpeed_TypeDef` enum, now they're all one place and written the same way, which makes using them easier. The other occurrences were found with grep and modified.  

Again, I hope this is acceptable.


cnlohr:
I thought I hit that function up but I guess not?

Issue #203, opened 2023-07-01
Baud Rate not being calculated


JimmusW:
Since the funconfig update, baud rate is skipping the calculation section in the header files.  It is being jammed into the BRR register directly.  This breaks the UART .  I can get around it by putting 416 in the define.  That was the value the old system calculated.  But it should probably be done right.

PR #204, opened 2023-07-01
fix uart printf baud

74th:
I tried running uartdemo, but it did not work because the baud rate was set directly on the BRR.
Since the CH32V003 has only one UART, I use the UART macro that was already there.


maxgerhardt:
So this fixes #203?

74th:
Yes !

PR #210, opened 2023-07-19
sysclk fixes and example project

hexeguitar:
This PR is a result of a few days of testing and digging into the clock system on the CH32V003.  
### Issues fixed:
- Wrong clock source for the PLL if the HSE is used.
- PA1 + PA2 reconfigured via AFIO to work with an xtal.
- leaves the HSI always on, which is required by the debug system to operate. It can be changed by changing one #define line.
### Other changes:  
**CSS  -Clock Security System** enabled. This allows a very easy unbricking (if normal way using programmer doesn't work) by shorting the PA1 (OSCI) to GND. HSE will fail, CSS will start the HSI and use it as sysclk source restoring the access via SWIO. The clock fail requires the NonMaskaleInterrrupt to clear the flag, otherwise it will end in the defaultHandler and an infinite loop. NMI IRQ has it's own declared alias handler which clears the flag. The user can still override it if there is more to do, like attempting HSE recovery, switching the PLL on.  
CSS can be disabled by declaring  
`#define FUNCONF_USE_CLK_SEC 0`
in the funconf.h file.

### New example project:
All the features can be tested using the included example project: **sysclk_config** which checks the registers and prints out via debug printf and UART (in case debug gets disabled) a summary of the clock system settings:  
```
System Clock Configuration Example
====================================
HSI Enabled
HSI trim = 16
HSI cal = 36
HSE Enabled
PLL Enabled
PLL Source = HSE
Clk Security Enabled
HSE bypass Disabled
Sysclk source = PLL 
MCO output: Disabled
```  
More details in the projects readme file.

I have tested all the clock config combinations on two boards: nanoCH32 and the WCH dev board.

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -1018,7 +1018,7 @@ void SystemInit()
 #endif
 
 #if defined(FUNCONF_USE_CLK_SEC) && FUNCONF_USE_CLK_SEC
-	#define RCC_CSS RCC_CSSON										// Enable clock security system
+	#define RCC_CSS RCC_CSSON									 	// Enable clock security system
 #else
 	#define RCC_CSS 0
 #endif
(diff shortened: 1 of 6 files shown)


eeucalyptus:
I just ran over the code and except for aestetics I think that this pr covers everything mine does and it adds functionality like the CSS and I found that my pr neglects HSE without PLL, so I will withdraw mine in favor for this one

Issue #212, opened 2023-07-25
3 Examples uncompilable


maxgerhardt:
https://github.com/cnlohr/ch32v003fun/actions/runs/5624816590

![image](https://github.com/cnlohr/ch32v003fun/assets/26485477/84d4b670-f18c-4548-bdee-875fbf490d6d)

maxgerhardt:
Fixed with 207010bebdec32d519667618162c011eef0ee129

PR #213, opened 2023-07-25
Issue #212 fixes - examples fail to compile

hexeguitar:
1. MCO example enabled both HSI and HSE, fixed by disabling HSE in funconf.h and enabling it in the example code.
2. SPI library still requires APB_CLOCK to be defined, which was removed in #208


cnlohr:
Thanks!

PR #215, opened 2023-07-25
Isochronous adc example

eeucalyptus:
I used [https://github.com/hyOzd/serialplot](https://github.com/hyOzd/serialplot) to test the serial output. Works fine with 48 kHz and 2MBaud, but if serial output is not required or the signal is downsampled before transmission, it should operate at the maximum possible sampling rate of 1 MHz. Useful for software oversampling and other DSP stuff :)


cnlohr:
Thank you!

PR #216, opened 2023-07-26
Disable DEBUGPRINTF I/O when FUNCONF_USE_DEBUGPRINTF = 0

emeb:
This patch will completely disable the SWDIO-based debug printf infrastructure when FUNCONF_USE_DEBUGPRINTF is set to zero. This allows printf() calls to be left in the project source without generating unwanted timeout delays when the debugger is not connected.


emeb:
Note - CI failures in examples/MCOtest and examples/spi_24L01_tx/rx don't appear to be caused by this patch.

cnlohr:
Thanks!

PR #217, opened 2023-07-30
Remove redundant debug shell scripts

emeb:
This PR removes several shell scripts that were inadvertently left in with example code when it was originally checked in. These scripts referred to external resources that are unlikely to be present for most users and should not be generally distributed.


cnlohr:
Thank you!

PR #220, opened 2023-08-01
Avoid bricking CH32V003J4M6 - For #219

### LINKED ISSUE #219: debugprintfdemo bricks CH32V003J4M6 since SWIO and PD4 share the same PIN
# Root Cause
The `SWIO` and `PD4` share the same PIN on CH32V003J4M6. In the `debugprintfdemo`, `PD4` is used for output, after flashing the chip is bricked since `SWIO` is taken as `PD4`.

# Solution
- The `debugprintfdemo` should avoid using `PD4`.
- The command `minichlink -ua` can unbrick the chip. However, if the chip was connected to the WCH-LinkE for a long time, it could be damaged. For me, one chip is unbricked, and another is too hot and damaged.

image

limingjie:
Avoid bricking CH32V003J4M6 by not using `PD4` in the demo.

### DIFF
--- a/examples/debugprintfdemo/debugprintfdemo.c
+++ b/examples/debugprintfdemo/debugprintfdemo.c
@@ -24,24 +24,20 @@ int main()
 	GPIOD->CFGLR &= ~(0xf<<(4*0));
 	GPIOD->CFGLR |= (GPIO_Speed_10MHz | GPIO_CNF_OUT_PP)<<(4*0);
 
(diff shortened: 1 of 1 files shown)


cnlohr:
I forgot that when I accepted the other PR.  Thank you for identifying this quickly, I will accept the revert now.

limingjie:
Thank you!

PR #222, opened 2023-08-01
tim2_encoder example based on tim2_pwm_remap

fr0sty1:
Uses partially remapped pins to read a rotary encoder (or any quadrature input) using pins that should be 5V tolerant


cnlohr:
I love the chart explaining things.

PR #226, opened 2023-08-11
[WIP] V10x/V20x/V30x support

AlexanderMandera:
`ch32v003fun.h` and `ch32v003fun.c` now have support for the other chips,
but everything is basically untested except for a blink sketch.

A blink example can be found in `examples_v20x` etc.
You can check the branch out and test it on your own hardware, let me know if you find errors or better solutions. 👍 

There are also some changes to minichlink in here but they are not completely ready yet.


cnlohr:
:)

PR #228, opened 2023-08-17
Add warning message if WCH-LinkE is detected in ARM mode

jeremyherbert:
See https://github.com/cnlohr/ch32v003fun/issues/227 for details.

### DIFF
--- a/minichlink/pgm-wch-linke.c
+++ b/minichlink/pgm-wch-linke.c
@@ -87,11 +87,18 @@ static inline libusb_device_handle * wch_link_base_setup( int inhibit_startup )
 	libusb_device *found = NULL;
 	ssize_t cnt = libusb_get_device_list(ctx, &list);
 	ssize_t i = 0;
+	int found_arm_programmer = 0;
 	for (i = 0; i < cnt; i++) {
 		libusb_device *device = list[i];
 		struct libusb_device_descriptor desc;
 		int r = libusb_get_device_descriptor(device,&desc);
 		if( r == 0 && desc.idVendor == 0x1a86 && desc.idProduct == 0x8010 ) { found = device; }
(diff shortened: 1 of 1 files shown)


minichlink/pgm-wch-linke.c
```diff
@@ -87,11 +87,18 @@ static inline libusb_device_handle * wch_link_base_setup( int inhibit_startup )
...
+	int found_arm_programmer = 0;
 	for (i = 0; i < cnt; i++) {
 		libusb_device *device = list[i];
 		struct libusb_device_descriptor desc;
 		int r = libusb_get_device_descriptor(device,&desc);
 		if( r == 0 && desc.idVendor == 0x1a86 && desc.idProduct == 0x8010 ) { found = device; }
+		if( r == 0 && desc.idVendor == 0x1a86 && desc.idProduct == 0x8012) { found_arm_programmer = 1; }
+	}
+
+	if (found_arm_programmer) {
```
eeucalyptus, line 99: Maybe it would be better to only output the warning if no risc-v programmer was found. I use a jlink for my arm chips, but I guess there are others who might use both versions in parallel? Would just be a move down by three lines

jeremyherbert: I originally had it there and moved it outside ;) My reasoning was that it's better to show the warning and let the user ignore it than have the issue possibly hidden. If they have a different RISC-V programmer connected simultaneously (such as one of the ESP32 based ones) then the warning would not show up if it was inside the if block.

eeucalyptus: Okay, your point is as valid as mine, I guess for a program that doesn't have the word "mini" in it's name, the solution would be to add a global verbose mode, but that's definitely out of scope for this tool. I'm fine with the warning being shown ;)

PR #235, opened 2023-09-08
Update standby examples and add implementation notes.

baoshi:
Set GPIOs to input pull-up/pull-down is required to achieve 10uA standby current. Update examples and added notes.

### DIFF
--- a/examples/standby_autowake/README.md
+++ b/examples/standby_autowake/README.md
@@ -6,8 +6,7 @@ This example serves to show how to put the CH32V003 into its lowest power state 
 
 Power consumption should be around 10uA.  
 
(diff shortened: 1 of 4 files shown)


cnlohr:
Really like some of these changes, especially the documentation ones, but, using large streams of binary digits is a recipe to confuse people, please use more of the defines from ch32v003fun.

marekmaskarinec:
Maybe this level of documentation could be the standard for all examples.

baoshi:
Have relooked into the standby current and found my previous result not entirely correct. Have updated GPIO configuration notes and examples accordingly.

cnlohr:
Looks pretty good now.

PR #238, opened 2023-09-15
fix the adc trigger settings

eeucalyptus:
The ADC_ExternalTrigConv_T1_TRGO was written to the wrong register. It didn't cause problems in the example, but as soon as you try to change the trigger.


cnlohr:
Excellent!

PR #245, opened 2023-09-30
more libs and examples

bitbank2:
This PR moved the ch32v_hal and spi_lcd color LCD libraries into extralibs and added a new example - Sensor_Platform with a bunch of I2C sensor libraries.


cnlohr:
Extralibs is only for libs that use the interfacing on the ch32v003.  Any device-specific libs go in the individual projects.

For something like the LCD, if you must, you can link directly to the code in the other example.

bitbank2:
I think that makes it harder for people to re-use common code that should be easily accessible. Maybe you could add a folder called "contrib" or "contrb_libs"? I hope there is another way to manage this.

PR #257, opened 2023-12-18
added __isenabled_irq __get_cpu_sp, minor linting

mrx23dot:
+white spaces


cnlohr:
While some of these are totally valid changes, (1) this codebase is tabs not spaces and (2) this is a LOT of churn.  I'm ok with some churn from time-to-time, but any kind of style changes or white space fixup needs to be separated out into separate PRs.

PR #258, opened 2023-12-21
examples/direct gpio: remove count variable

amitesh-singh:
count variable does not do anything


cnlohr:
This is one wholesome pr.

PR #259, opened 2023-12-21
fix compilation for arch linux

amitesh-singh:
on arch linux, the risc compiler name is different. Add some logic to pick correct risc compiler on arch linux

### DIFF
--- a/ch32v003fun/ch32v003fun.mk
+++ b/ch32v003fun/ch32v003fun.mk
@@ -1,5 +1,12 @@
-
-PREFIX?=riscv64-unknown-elf
+ifeq ($(OS),Windows_NT)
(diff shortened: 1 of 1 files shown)


cnlohr:
Can you change it so you test if the typical compiler exists, else use a different one? I.e.

https://stackoverflow.com/a/25668869/2926815

When not on Windows (Windows doesn't have which)

cnlohr:
*poke* Any chance you could update this to do the check for the if-exists?

cnlohr:
*poke* @amitesh-singh Any chance you could update this to do the check for the if-exists?

amitesh-singh:
@cnlohr updated. Thanks for the reminder.

cnlohr:
Thanks!

PR #260, opened 2023-12-23
Fix typo preventing specpgm hint from being read on cmdline

BobRyan530:
This fixes a typo preventing programmers to be specified on the command line. It also adds a udev rule for the rv003usb bootloader.


cnlohr:
Thanks for fixing these.

PR #263, opened 2024-01-04
Fix comment

tintinnabulate:


cnlohr:
how is the reference redundant?

tintinnabulate:
> how is the reference redundant?

The reference in the comment is to PORTC pin 6, but `WS2812BSimpleSend(..)` takes `port` and `pin` as parameters. Hence the reference is redundant.

cnlohr:
OH MY GOSH.  I am so blind.  Thank you for this PR.

PR #279, opened 2024-02-18
Fixup PlatformIO CI

maxgerhardt:
* generates the used linker script file with the `gcc` preprocessor exactly like the Makefile does
* instead of adding `examples/<name of environment>` to the include path (to find `funconfig.h`), the wanted include folder is found automatically by a sidechannel in the `build_src_filter`, which is set for every example
  * Otherwise, building `examples/blink` and `examples_v10x/blink` would not be possible because the previous method would have required both environments to be named `[env:blink]`
* now also uses one more line per example that determines whether to use the 003, 10x, 20x or 30x envrionment
* Completely add every single example (for 003, 10x, 20x, 30x) to the `platformio.ini`
* Update the instructions for adding new examples
* removes pregenerated LD file


maxgerhardt:
As a bonus, fixes Makefile CI to now also build V10x, V20x and V30x examples, which were not built before. See e.g. `examples_v10x/blink`.

PR #281, opened 2024-02-21
added acronym definitions and helpful links to README

BebeSparkelSparkel:
I'm new to microcontrollers and these acronym titles and links were helpful to me.

Help needed with EVT acronym.


cnlohr:
I think it means something like "Evaluation/Test" but why the acronym worked its way into the software stack, I have no idea.

cnlohr:
I would just not worry about it and treat it as a nonsense acronym.

BebeSparkelSparkel:
Just added the approximate definition you suggested.

BebeSparkelSparkel:
@dwillmore As someone not  in the industry and just looking at the intro it is hard reading to lookup the numerous acronyms and determine which one is applicable.

Also, I think there is a hobby of interest in this chip since EEVBlog posted about it https://www.eevblog.com/forum/blog/eevblog-1524-the-10-cent-risc-v-processor-ch32v003/

cnlohr:
Conflicts now anyway. We might add a few of these here and there in the future.

PR #282, opened 2024-03-01
Add minichlink unbrick to makefile

MartinEssink:
Add the minichlink unbrick command to the makefile, such that the chip can be unbricked using 'make unbrick'.


cnlohr:
thanks

PR #287, opened 2024-03-06
Make udev rules work on distros without plugdev group

vinicentus:
The `plugdev` group is from what I understand an Ubuntu specific thing. This fixes the udev rules so they work on other distros such as Arch. I also corrected the name of the udev rules file in the makefile.

Tested on Arch.

Let me know if this looks ok.


dwillmore:
plugdev is a systemd thing, so any distro using it: Debian, Ubuntu, Fedora, SUSE, CoreOS, etc.  This change will break udev for the majority of Linux users.

Please reconsider this change.

A reasonable change would include a test for the distro of the machine by testing the ID field of /etc/os-release.

cnlohr:
I agree with @dwillmore here.  If your distro is missing plugdev, add it.  Otherwise, all users, including users you may want to have limited (non-USB access) can get access to it.

cnlohr:
It's pretty easy to say:

```sh
sudo groupadd plugdev
sudo sudo usermod -aG plugdev YOUR_USERNAME
```

We could document that.

dwillmore:
Yep, just add it to the wiki.

dwillmore:
Seems like the Makefile needs to be fixed at least.

cnlohr:
We definitely do not want to allow `uaccess` as that could definitely be a source of CVEs moving forward. We can add a check to minichlink to see if the current user is in the `plugdev` group, and if not, warn them?

cnlohr:
I have added a check for plugdev in minichlink, as well as added a note on the installation instructions. https://github.com/cnlohr/ch32v003fun/wiki/Installation#archmanjaro

vinicentus:
That seems like a good solution. Please also update the makefile to install the correct udev riles file: https://github.com/cnlohr/ch32v003fun/blob/master/minichlink%2FMakefile#L43

Also, I think `udevadm --reload` and optionally `udevadm trigger` is a more universal way to reload the rules.

cnlohr:
Fixed.

vinicentus:
Amazing, thanks!

Actually it should be `udevadm control --reload` for it to work. I was typing my last message from memory on my phone :D

I tested with that modification and by adding the plugdev group and it works.

But, I can't get the minichlink plugdev group warning to trigger. How is it supposed to work? Shouldn't I see the message when trying to run any minichlink command when there is no such group or when I'm not in the group?

CodeAsm:
> plugdev is a systemd thing, so any distro using it: Debian, Ubuntu, Fedora, SUSE, CoreOS, etc. This change will break udev for the majority of Linux users.
> 
> Please reconsider this change.
> 
> A reasonable change would include a test for the distro of the machine by testing the ID field of /etc/os-release.

Am i missing something? https://wiki.archlinux.org/title/Udev#Allowing_regular_users_to_use_devices
seems to fix this. and I believed reading before, that plugdev is definitly a "Ubuntuism" thing

"plugdev" is considered a bug:
https://bugzilla.redhat.com/show_bug.cgi?id=815093
``The group "plugdev" is a pretty misguided and inflexible Ubuntu concept to grant
device access to ordinary users.`` So while i wait for riscv64-unknown-elf-gcc etc to compile, wondering if I can change this plugdev behavior.

prosper00:
just my 2c, but i feel like sysadmin stuff - such as ensuring permissions - is my own responsibility. And how i choose to implement it - via groups, policies, udev, etc - is my job. Applications should simply assume that I've done my job correctly, and display an error -maybe with a hint- if i haven't. But it bugs me when they decide to enforce a particular assumption about how a thing must be done.

dwillmore:
Agreed.  There's info on the wiki about how to install things on various
operating systems.  The code should be able to assume that people followed
the directions on the can.

cnlohr:
> Well yeah, arch linux also considers plugdev a bug. But I definitely get the concerns of automatically -perhaps without the user knowing about it- allowing all users on a system access to a physical device.

People use hidapi.  They have to with it.  You can only plug your ears and yell "I can't hear you" for so long.  If a user has adm, then it's just silly.

PR #288, opened 2024-03-06
Add GPIOB define for CH32V10x, CH32V20x and CH32V30x

renzenicolai:
A tiny change but a very useful one if you want to use port B :smile:


cnlohr:
Smol Port B.  As gift.  Enjoy.

PR #297, opened 2024-03-31
Fix linker related error.

imbus64:
These patches fixes a linker issue where the .note.gnu.build-id would corrupt the binary. Some other commonly discarded sections are also thrown in, some of which have a measurable effect on the final binary. The pre-built binary present in the blink example directory was also removed.


cnlohr:
@imbus64 do not delete the .bin.  It is there on purpose so people can flash it, I want the rest of the PR.  Please update.

imbus64:
@imbus64 Done.

PR #322, opened 2024-06-13
Add UART DMA TX example

doragasu:


cnlohr:
@doragasu this looks good - but you need a README.md that describes what it is you are doing, and what pins to connect, etc.

doragasu:
Will add one then when I get some time

cnlohr:
Also, you will need to rebase with the new CI rules (just FYI).

PR #326, opened 2024-06-22
Add sh1107

cnlohr:


emeb:
I've tested this against all three supported SSD1306 OLEDs and it doesn't appear to have broken anything. I don't have SH1107 displays so I can't verify if that's working properly.

PR #335, opened 2024-07-06
Remove plugdev warning on OSX

cHemingway:
On OSX, minichlink shows the warning 
> WARNING: You are not in the plugdev group, the canned udev rules will not work on your system

But udev does not exist on OSX, so this warning can be removed. I just extended the trick that makes it not appear on Windows either.


cnlohr:
Thanks.

PR #341, opened 2024-07-11
fix compiler warnings

20dka:
fixes for `[-Wsign-compare]` and `[-Wunused-parameter]`


20dka:
with this, should `[-Wextra]` be added to the base Makefile?

cnlohr:
Looks good.

And let's not worry about it now.

PR #349, opened 2024-07-14
Better PlatformIO documentation, build X035 examples too

maxgerhardt:
* Fix linkerscript generation for CH32X035 in accordance to the Makefile
* Add all x035 examples to the `platformio.ini` so that they will be compile-checked in the CI 
  * along with minor adapation of `add_includes.py` to account for build filters that point to one specific C file (otherwise, breaks usb example with `fsusb.c` being built twice, once as .c file once as inclusion of fsusb.h) 
* Revamp PlatformIO documentation
  * Strongly advise people to use the "New Project" wizard which newly gained native `framework = ch32v003fun` support per https://github.com/Community-PIO-CH32V/platform-ch32v/commit/31cdd721af55ca35cc0787440fcda610471d4a9d
  * Accounts for feedback in #329 and other examples of people creating the projects in a very wrong and weird way, or not finding some key buttons like the project environment switcher
* sneaky: add a `package.json` to this repo. This is a package metadata file hard-required by PlatformIO. With this file in this repo, it is possible for PlatformIO projects to directly reference `platform_packages = framework-ch32v003fun@https://github.com/cnlohr/ch32v003fun.git` ([docs](https://docs.platformio.org/en/latest/projectconf/sections/env/options/platform/platform_packages.html)) for the latest version. If unwanted, I will just keep this file in my fork.


cnlohr:
Thank you for doing this.

PR #351, opened 2024-07-15
Fix bug where SSI is set by the SPI_Mode_Master mask

fsaev:
0x0104 is bitmask 100000100 which sets bit 2 (MSTR) and bit 8 (SSI). The latter has no relevance to master mode.

### DIFF
--- a/ch32v003fun/ch32v003fun.h
+++ b/ch32v003fun/ch32v003fun.h
@@ -11161,7 +11161,7 @@ typedef struct
 #define SPI_Direction_1Line_Tx             ((uint16_t)0xC000)
 
 /* SPI_mode */
-#define SPI_Mode_Master                    ((uint16_t)0x0104)
+#define SPI_Mode_Master                    ((uint16_t)0x0104) /* Sets MSTR, as well as SSI, which is required for Master Mode */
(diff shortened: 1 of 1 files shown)


fsaev:
Is this a workaround for a peripheral bug? I could change this to a comment instead.

cnlohr:
All of the documentation from OpenWCH points at master requiring SSI.  https://github.com/search?q=repo%3Aopenwch%2Fch32v003%20SPI_Mode_Master&type=code This could be an error on their part (And it wouldn't be the first).  Can you test it without SSI and see?

cnlohr:
Right now, after digging through, I am going to recommend we leave it as-is, unless there is a very specific reason why a user would _need_ to set SSI low while in master mode.

fsaev:
That sounds reasonable. Should I maybe replace it with a comment to describe what is going on?

cnlohr:
yes! That would be wonderful!  Just note that the official EVT always sets SSI when in master mode, so we have decided to mimic the behavior.  I am happy to accept those sorts of clarifications as PRs!

fsaev:
How does it look?

cnlohr:
Looks good.

PR #352, opened 2024-07-15
Name fix

fsaev:
This fixes some misleading name choices where the GPIO speed is referred to as maximum 50 MHz, while it is in fact 30 MHz on the CH32V003.

For GPIOSpeed_TypeDef I added explicit hexadecimal values to the enum, as implicit values are undefined and also for the sake of consistency and clarity.

I've also done a pedantic correction of "Mhz" to "MHz", because I think concise naming is valuable in a library


cnlohr:
(1) Please leave it as 50MHz to match the rest of their line of products.  And this will help keep all the ch32x00x processors in line.  There's also a lot of documentation calling it 50, so switching to 30 would be confusing.

image


(2) Don't include the change to `SPI_Mode_Master` until we are confident about its intended behavior.

fsaev:
I'm looking at the reference manual for the CV32V003 (CH32V003RM-EN v1.7)
Screenshot_20240716_185458

I understand that it's tedious to change naming in a library. Maybe we should consider a major release that can allow for breaking changes?

cnlohr:
No.  We should not update. The reference manual is just changing their opinion of where the cutoff for the lowpass occurs.  IMO, this should not be modified in code definitions.  I will ping others on the Discord to see what they think.

cnlohr:
Additionally, this makes it so code written for the 003 is now no longer directly portable to the 203, and for no particular reason.

cnlohr:
Ok, did a little more legwork.  It is 50MHz drive.  Not sure why the reference manual changed it.

cnlohr:
Ok, did a little more legwork. It is 50MHz drive. Not sure why the reference manual changed it.
image

fsaev:
Alright! I will make a new pull request that only fixes the explicit enum values

cnlohr:
Sorry, I know that was a lot of work to end up back where we started :-/

PR #354, opened 2024-07-16
Add example for using SPI TX with DMA on the CH32V003

fsaev:
Implement SPI TX  using DMA, with 8-bit mode and soft NSS using a small state machine. This example prints the sequence of 0 .. 7 repeatedly.

### DIFF
--- a/examples/dma_spi/Makefile
+++ b/examples/dma_spi/Makefile
@@ -1,0 +1,11 @@
+all : flash
+
+TARGET:=dma_spi
+TARGET_MCU:=CH32V003
+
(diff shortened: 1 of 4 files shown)


cnlohr:
You will need to include a README.md that details how to use this feature, how you would connect to the chip to use it and a high level description of what your code does.  For instance, include a note about which pin on the chip is configured for TX, etc.

fsaev:
You can have a look now.

cnlohr:
This looks great! Will run against CI and accept if it passes.

PR #356, opened 2024-07-18
Added target for object file

niansa:
Useful for use in bigger non-makefile (cmake/meson) projects


niansa:
Sorry, GitHubs web IDE ruined that tabulator

cnlohr:
Would you mind adding a comment explaining that it's not used in the bare-bones tree but can be used in the other applications you were thinking?

cnlohr:
Err I can just accept and add.

cnlohr:
Fixed in main.

PR #362, opened 2024-07-27
Support TIM3 on CH32X03x

jurriaan:
Noticed that a few defines where missing. This prevented the usage of TIM3 on my CH32X035


cnlohr:
@jurriaan  Have you tested it? Does it work?  Do you mind publishing an example?

cnlohr:
something like the other examples in the examples folder for tim1 and tim2, except it would be in the [examples_x035](https://github.com/cnlohr/ch32v003fun/tree/master/examples_x035) folder

PR #363, opened 2024-07-30
Add TIM3 PWM example for CH32X035

jurriaan:
Adds a PWM example for the general purpose TIM3 timer for the CH32X035.


cnlohr:
This looks great!!! Thank you!

PR #366, opened 2024-08-01
Fix I2C_CKCFGR_CCR typo for slow clock case

emeb:
An ancient untested typo in the name of an I2C configuration register was corrected and has been tested to compile correctly.


cnlohr:
Thank you!

PR #368, opened 2024-08-06
Break out random number generation to separate header

emeb:
This change replaces the original LFSR code with a somewhat more efficient approach and pulls the function out into a separate header file that could be more easily used by others. It also adds two wrapper functions that emulate the behavior of the rand() and srand() functions from <stdlib.h>.

### DIFF
--- a/examples/spi_oled/rand.h
+++ b/examples/spi_oled/rand.h
@@ -1,0 +1,51 @@
+/*
+ * rand.h - LFSR-based random number generation
+ * Released to Public Domain 08-05-2024 E. Brombaugh
+ */
+ 
+/*
+ * Common function
+ * seed = 0 - advances LFSR state and returns result
+ * seed = anything else - set seed to that value
+ * bit = 1-32 - number of bits to generate
+ */
+uint32_t rnd_fun(uint32_t seed, uint8_t bits)
+{
+	static uint32_t rnd_state = 1;
+	uint32_t lsb;
+	
+	if(seed)
+	{
+		rnd_state = seed;
+		return 0;
+	}
+	
+	while(bits--)
+	{
+		lsb = rnd_state & 1;
+		rnd_state >>= 1;
+		if(lsb)
+			rnd_state ^= 0x80200003;
+	}
+	
+	return rnd_state;
+}
+
+/*
+ * stdlib random emulation
+ */
+
+#define RAND_MAX 0x7FFFFFFF
+
+/* generates 31 bits PRN */
+int rand(void)
+{
+	return rnd_fun(0,31)&RAND_MAX;
+}
+
+/* set seed */
+void srand(unsigned int seed)
+{
+	rnd_fun(seed,0);
+}
+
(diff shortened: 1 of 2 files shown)


mengstr:
Maybe I'm just oldskool (and also an old man yelling at the clouds), but it really grinds my gears when I see actual code and not just declarations and such in header files.  But it seems to become more common as time passes by - does this habit come from Arduino or some other language?

While implementing a RND() for my SUBLEQ TinyBasic I (re-)discovered how horrible LFSR random generators are - they emit long runs of the same bits, regardless of where you mask and shift the result/seed. So I ended up with mixing the outputs of a LFSR and and LCG generator. https://github.com/mengstr/SUBASIC/blob/main/RANDOM.inc

emeb:
"Single File Headers" that contain both code and definitions are definitely a choice. I don't use them in my own projects, but they're the style that has been chosen for ch32v003fun.

mengstr:
Is it really? If I look at Charles code here including many of the original examples and the minichlink virtually all code is in the C files, and the headers are #defines,  function declarations and a very few inlines functions and macros. 

But I'm not sure there's even one example that are using multiple C files here. It took me a while to figure out how to get this running when using the makefile.mk from the code in my own projects - especially getting the include paths correct. So maybe it's a good thing to have code in the h-files in order to make things easier.

emeb:
Back when I started contributing to this project we were specifically asked to lean into single file headers for any code that could be shared. Most of the examples don't actually have a lot of stuff that really needs to be broken out like that, but over time many of those that do have been moved into the extralibs area and all that code is SFH. In this case, we had some discussions on Discourse about moving the rand() function into the main ch32v003fun library so others could access it easily and this PR is leading up to that.

WRT "quality" of the random bits - it's a trade-off between goodness and the number of cycles you want to burn and tends to be a fairly contentious discussion with a lot of opinions about the right way to do things. This approach is one that I've been comfortable with for a while and I've run some tests on this specific case to ensure it really does produce 2^32-1 unique states before repeating, but if anyone else has a favorite that works better then please put it out there for everyone.

cnlohr:
We are deliberately experimenting with leaning more towards code-in-header, though, I'm def down for having definitions at the top of the file.

There is little reason now to separate out the two.  But, I admit it is more of an experiment to see how it goes.  So far, I really like how things have gone with that. 

That said, the output of this rand function is a little weird because it cascades.  As in there is correlation between samples. Which for a LOT of applications is just fine.

emeb:
Yes, because of the shifting there is correlation if you use more bits than are requested by setting the bits parameter in the common call. To keep compatibility with the libc function and eliminate correlation the rand() wrapper generates 31 bits and masks off the msbit, and if you need only 8 bits (for example) then call the common function with bits set to 8 and use only those lsbits.

cnlohr:
Oh, you know, I didn't put 2 and 2 together til now re: that.  That is very obvious and very slow.   But I guess it is what it is.

Issue #303, opened 2024-04-13
v30x blink example don't run if TARGET_MCU_PACKAGE sets to CH32V307VCT6


romainreignier:
Hi,
First, I would like to thank you for this amazing project.
I have successfully tried the blink example from the `examples_v30x` on a `CH32V307 EVT by SCDZ` board.
But the board is actually populated with a `CH32V307VCT6` and the Makefile selects a CH32V305RBT6, I have then replaced it with the following:

```
all : flash

TARGET:=blink
TARGET_MCU:=CH32V307
TARGET_MCU_PACKAGE:=CH32V307VCT6

include ../../ch32v003fun/ch32v003fun.mk

flash : cv_flash
clean : cv_clean
```

And after a `make clean` and then `make`, the code does not run once uploaded to the board.

I have also tried with the `debugprintfdemo` copied from the V003 examples, it works with the V305RBT6 settings but not with the V307VCT6.

From the `ch32v003fun.mk`, the difference is only the `MCU_PACKAGE` and the sizes for the FLASH and RAM.

I have then tried to configure `TARGET_MCU_PACKAGE:=CH32V307RBT6` and in that case, the code runs.

When it does not work, I have tried to use GDB and adding a breakpoint on the first line of the main on the `SystemInit()` line. The breakpoint is reached but on first step, I am in the `DefaultIRQHandler`.

Does anyone have a clue of why I have this behavior when using the config matching my MCU?

cnlohr:
Sadly, at the moment no one is looking into this specific target.  If you are up for seeing where we've made a mistake, it would be great to let us know. 

There's plarform-specific code in the .ld, Makefile, and ch32v003.h

As a note, when changing the .ld, you will need to remove the auto-generated .ld for your part in the ch32v003fun folder.

Please let us know what you find.

cnlohr:
@AlexanderMandera - do you recall if you ever tested with CH32V307VCxx?  Should this line read 2 instead of 1? 			MCU_PACKAGE:=1 https://github.com/cnlohr/ch32v003fun/blob/6c3c515598db6d3d50f8087da2f694ee417fc445/ch32v003fun/ch32v003fun.mk#L114C14-L114C18

@romainreignier Can you test with changing that line of that file from 2 to 1, removing the generated .ld file, and seeing if it works then?

romainreignier:
[...]

Issue #280, opened 2024-02-18
Configurable RAM sizes for V20x, V30x


maxgerhardt:
The linker script

https://github.com/cnlohr/ch32v003fun/blob/6c14912e79adf9097d46c3f4688da3b567d64d6c/ch32v003fun/ch32v003fun.ld#L31-L40

for V30x only gives the "256K Flash + 64K RAM" and "128K Flash + 32K RAM" option. This is however wrong: If the 128K Flash option doesn't exist naturally. The options for 30x are

![grafik](https://github.com/cnlohr/ch32v003fun/assets/26485477/a26aa99b-959a-4554-946b-45860f5d75e9)

And for 20x

bmx:
![image](https://github.com/cnlohr/ch32v003fun/assets/622675/722e3171-baed-43a9-a21c-54399f2ed323)
Who's wrong? the datasheet? the linkutility?

cnlohr:
> Through invocation of the blackest dark magic I've ever seen, flash cells become RAM cells, and their sum is a constant.

wat

Can you validate the actual performance?

maxgerhardt:
> Who's wrong? the datasheet? the linkutility?

Ah, indeed I'm wrong, or rather, confusingly worded: The V30x devices marked "256K Flash + 64K SRAM" parts are actually configurable for the 4 options shown in the link utility. This is only the CH32V307(RC/WC/VC) and CH32V303(RC/VC) chips. For the other chips, 303(CB,RB), 305(FB,RB), they **do** have 128K flash and 32K RAM *fixed*.

![grafik](https://github.com/cnlohr/ch32v003fun/assets/26485477/921c1d88-1661-43dd-b74f-0d2ff93777df)
![grafik](https://github.com/cnlohr/ch32v003fun/assets/26485477/4e74e1a8-1f01-4a8c-997e-18887ab4b75f)

For V20x series chips, only V203RB has configurable sizes from selections, all others are 32K+10K or 64K+20K.

The point still stands however, V30x sreies chips can actually have 5 different configurations, the `.ld` file makes only 2 possible.

cnlohr:
@dwillmore has been postulating that the flash is in fact RAM, and that it's somewhat arbitrary how it's configured. I am extremely curious if we find something like that.   It seems that there are bits in the option rom for configuring the bits, see `FLASH_OBR`
[...]

Issue #232, opened 2023-08-23
minichlink resetting option bytes on flash operations


ddrown:
I'm using the ch32v307, and it has a register to determine the ram/flash split. I'm guessing it's doing a shadow copy of the flash contents into ram. This is the option byte at 0x1ffff802, which is copied to the register OBR in the flash controller at startup.

I can set the option byte to use a 192KB flash/128KB sram split just fine with:
```
    FLASH_Unlock();
    FLASH_ProgramOptionByteData(0x1FFFF802, (FLASH_GetUserOptionByte() & 0b00111111));
    FLASH_Lock();
```

The board picks up my change through a reset, by pushing the reset button or using `minichlink -a -b` or `NVIC_SystemReset();`

Reading back the option bytes:

```
volatile uint8_t *optionBytes = (uint8_t *)0x1FFFF800;
printf("USER=%x %x %x %x\r\n", optionBytes[0], optionBytes[1], optionBytes[2], optionBytes[3]);
```

I get the expected value:
```
USER=a5 5a 3f c0
```

And the flash OBR register has the split I want:
```
OBR = 0x3fffcfc (192K+128K RSVR1 RSVR2 RSVR3 STANDBY_RST STOP_RST IWDG_SW)
```

But when I flash a new image with `minichlink -a -w build/example.bin 0x08000000 -b`, something is resetting my user option bytes back to 0xff:

```
USER=a5 5a ff 0
```

Which resets back to the 288K/32K memory split:

```
OBR = 0x3fffffc (288K+32K RSVR1 RSVR2 RSVR3 STANDBY_RST STOP_RST IWDG_SW)
```

I experimented with commenting out OBKEYR unlocking in `InternalUnlockFlash`, but that didn't change this behavior.

Any ideas on what to try next?

ddrown:
wlink doesn't work properly with my WCH-LinkE:

```
$ wlink flash --address 0x08000000 build/example.bin
17:01:44 [INFO] WCH-Link v2.9 (WCH-LinkE-CH32V305)
17:01:44 [INFO] Attached chip: CH32V30x(0x30700528)
17:01:44 [INFO] Read build/example.bin as Binary format
17:01:44 [INFO] Flashing 9792 bytes to 0x08000000
17:01:44 [INFO] Flash done
17:01:45 [INFO] Now reset...
17:01:45 [INFO] Resume executing...
Error: DMI Status Failed
```

[...]

PR #369, opened 2024-08-08
Configurable flash / ram split with minichlink

iracigt:
This adds support for changing the amount of SRAM available in parts where the flash / SRAM split is configurable. Right now, this only works for the wch-linke adapter. The split is given as a command line flag and applied during the bulk memory write. Without this, the CH32V307 is limited to 32k of SRAM of a possible 128k. Adding this setting addresses issues #232, #280, and (in part) #303.

As referenced in issues #232, #280, and #303, some of the CH32V20x and  CH32V30x parts have a user configurable option trading flash size for user RAM size. As described by the CH32V307 datasheet (pg 4):
> The products with 256K FLASH+64K SRAM support user select word to be configured as one of several
combinations of (192K FLASH+128K SRAM), (224K FLASH+96K SRAM), (256K FLASH+64K SRAM),
(288K FLASH+32K SRAM).

Speculation is that the chips cannot execute from flash and instead copy the entire contents of the firmware image to hidden internal SRAM. This flag would then be configuring the amount of that internal RAM used for code storage vs user RAM.

This information is stored in the user option section of flash. According to #232, the official programmer sets this via a special command, the same one used to set readout protection. This patch uses the command described there, however does sanity checks based on chip id and applies the setting only if appropriate for the target.

I suspect there is a more general solution to be had here. The flag can be set by firmware running on the chip. Updates require unlocking the flash, making the write, and then become active after a reset. One generic possibility would be to upload a code stub that sets the flag. Another would be to use the read word and write word functionality to set it.

Another drawback to this patch is the setting is only applied during the bulk memory write. This is fine for the common case of uploading code. One weirdness is that the new `-S [FLASH kB] [RAM kB]` flag needs to appear before the `-w` flag. Note that the current erase and program workflow clears this flag to the default (32k SRAM for the '307). I'm open to suggestions on how to improve that aspect.

I've tested this patch with my WCH [CH32V307V-EVT](https://www.lcsc.com/product-detail/Development-Boards-Kits_WCH-Jiangsu-Qin-Heng-CH32V307V-EVT-R1_C2943980.html) dev board and its integrated WCH programmer. My testing has been somewhat limited, but an application that allocates a 48k SRAM buffer and uses it seems to be working as expected. I don't currently have other hardware to test on.

### DIFF
--- a/ch32v003fun/ch32v003fun.h
+++ b/ch32v003fun/ch32v003fun.h
@@ -789,6 +789,19 @@ typedef struct
 	__IO uint32_t RES3;
 } ESG_TypeDef;
 
+typedef struct
+{
+    union
+    {   
+        __I uint32_t CHIPID;
+        struct 
+        {
+            __I uint16_t REVID;
+	        __I uint16_t DEVID;
+        };
+    };
+} INFO_TypeDef;
+
 #if defined(CH32V30x)
 /* FSMC Bank1 Registers */
 typedef struct
@@ -2483,7 +2496,7 @@ typedef struct
 
 #define OB_BASE                                 ((uint32_t)0x1FFFF800)    /* Flash Option Bytes base address */
 #define ESIG_BASE                               ((uint32_t)0x1FFFF7E0)
-#define CHIP_ID                                 *((uint32_t*)0x1FFFF704) // Mentioned in ch32v30x_dbgmcu.c, may not work on all processors.
+#define INFO_BASE                               ((uint32_t)0x1FFFF704)
 
 #if defined(CH32V003) || defined(CH32V10x)
 #define EXTEN_BASE                              ((uint32_t)0x40023800)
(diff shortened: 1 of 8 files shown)


cw2:
> Speculation is that the chips cannot execute from flash and instead copy the entire contents of the firmware image to hidden internal SRAM.

Sorry for hijacking the PR, I just recalled this has actually been confirmed - https://cpldcpu.wordpress.com/2024/05/01/decapsulating-the-ch32v203-reveals-a-separate-flash-die/

cnlohr:
Oh no, I already have something like this in another one of my branches.  But I do it in the Makefile.  I have it working with the 307. 

Sorry I just noticed this now.

This is much more comprehensive than my change though.  Blahhh I will see if I can find common ground.

eeucalyptus:
Could you rebase the branch to the current upstream branch? 
Unfortunately I don't have a controller with configurable memory for testing. Are you on the discord? Could you maybe hmu for testing? handle is the same as here

iracigt:
Rebase done. I'm not currently in the discord, but you can send an invite link to my email in the `git log` if you like.

It's probably worth restructuring how this patch works. The current approach feels inconsistent with the rest of minichlink. Ideally setting the split would be its own operation. Right now it just changes the behavior of the flash erase (?) sequence of the flash write function. I'm working now on a version that uses `ReadWord` and `WriteWord` instead, which should be compatible with all programmers, not just the WCH-LinkE.

cnlohr:
Invite sent.  I'd like to chat before we merge this.

iracigt:
I've rewritten my code in minichlink to be more generic by using `ReadWord`, `WriteWord`, `ReadHalfWord`, and `WriteHalfWord` in place of the wch-linke specific operations. This required reading the CHIP_ID, so I added a register definition for that to `ch32v003fun.h`. There's also test program in examples that prints the current settings and probes RAM until it reaches the end and faults.

Overall the `-S [FLASH kB] [RAM kB]` flags are still there, but now are a separate operation. This means you should pass them after `-w` if using it while writing a new flash image. It can also now be used standalone to change the split without erasing flash.

Issue #371, opened 2024-08-09
Minichlink stuck in an infinite loop with WCH-LinkE and some CH32V003


mverdon:
I've just received a new batch of custom PCBs with 003s on them and I've been trying to probe each of them with my WCH-LinkE WCH-LinkE-R0-1v3 with firmware 2.12. I always encounter the same issue when running `minichlink -i` never exits but probing my old batch works just fine.

The difference I see just by looking at the chips is that the batch/date code differs on the packages. It shows `WCH V003F4 02D13` on the old ones and `WCH V003F4 12E22` on the new ones. These are supposed to be the same part number CH32V003F4U6. It could be nothing but may explain the difference in behaviour.

I've been trying to debug the issue and it looks like the problem starts when reading the register `DMHARTINFO` in `StaticUpdatePROGBUFRegs()`. The correct value `0x002120f4` is read on the old batch but somehow the new batch always returns `0xffffffff`. After that, since it cannot retrieve the correct offset for DATA0, the whole process fails then it gets stuck in `WaitForDoneOp()` where reading `DMABSTRACTCS` always gives `0xffffffff`.

Here are the logs for the old batch where it correctly probes the chip:
```
Found WCH Link
WCH Programmer is LinkE version 2.12
Full Chip Type Reply: [8] 82-0d-05-09-00-31-05-00
Detected: CH32V003
LEWriteReg32 10, 80000001
LEWriteReg32 10, 80000001
LEWriteReg32 10, 80000003
LEWriteReg32 10, 80000001
LEWriteReg32 16, 00000700
LEWriteReg32 18, 00000000
LEWriteReg32 17, 00221000
LEReadReg32 16, 08000002
Setup success
Flash Storage: 16 kB
Part UUID    : 29-0f-ab-cd-90-fc-bc-20
PFlags       : ff-ff-ff-ff
Part Type (B): 07-13-bb-91
Read protection: disabled
Interface Setup
LEWriteReg32 7e, 5aa50400
LEWriteReg32 7d, 5aa50400
LEWriteReg32 7d, 5aa50400
LEWriteReg32 10, 80000001
LEWriteReg32 10, 80000001
LEWriteReg32 18, 00000000
LEWriteReg32 20, 40044180
LEWriteReg32 21, c1040411
LEWriteReg32 22, 9002c180
LEReadReg32 12, 002120f4
LEWriteReg32 04, e00000f4
LEWriteReg32 17, 0023100a
[...]

PR #377, opened 2024-08-15
Prevent minichlink from hanging with some CH32V003 revisions and WCH-LinkE

mverdon:
This is a fix for issue #371. Some CH32V003 revisions will get minichlink stuck in an infinite loop when writing to `DMCFGR` twice in a row.

If anyone has chips with revisions different from what is listed in the issue (i.e. `02D13` and `12E22`) and different from the SOP-8 variant with revision `D26`, please confirm that this does not break minichlink.


cnlohr:
I just can't find a situation where this fails so I'll just merge.

danielkucera:
To add more info here, the flashing also fails with official WCH-LinkUtility V2.30 with following error:
```
12:02:14:354>> The choosen WCH-Link is:[#1]
12:02:14:369>> Begin to set chip type...
12:02:14:470>> Succeed
12:02:14:470>> Begin to Program...
12:02:14:470>> Failed,please check the status of WCH-Link and the chip! Error:4
12:02:14:470>> Operation is Failed
```

Snímka

Reads work ok though.

Thank you for the great work on this project!

xsrf:
Thx, this probably saved me so much time! I just got a complete batch of PCBs and none would work. minichlink identified the chip but wouldn't upload. This saved me. The chips I've used before without issues were `J4M6 D03` and `J4M6 E11`. The chips that now caused the issue with my old minichlink were `J4M6 E36`.
I checked, and my old chips also work fine with the new minichlink version. However, the "Part Type:" reports different values and also `minichlink` and `minichlink -i` now cause a reboot of the chip which my old version didn't, but that's a non issue and probably not even related to this pr.
I used the windows binary from this repo.

gsteiert:
This saved me too. Thanks for fixing this.

PR #379, opened 2024-08-17
Weak linkage for symbols that might conflict with libc

niansa:
This change allows me to completely toss my libc function internalization hack :+1:

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -88,7 +88,9 @@ void __libc_init_array(void)
 #include <stdint.h>
 #include <ch32v003fun.h>
 
-int errno;
+#define WEAK __attribute__((weak))
+
+WEAK int errno;
 
 int mini_vsnprintf( char *buffer, unsigned int buffer_len, const char *fmt, va_list va );
 int mini_vpprintf( int (*puts)(char* s, int len, void* buf), void* buf, const char *fmt, va_list va );
@@ -101,7 +103,7 @@ static int __puts_uart( char *s, int len, void *buf )
 	return len;
 }
 
-int printf( const char* format, ... )
+WEAK int printf( const char* format, ... )
 {
 	va_list args;
 	va_start( args, format );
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
Since this PR introduces this macro:

```
#define WEAK __attribute__((weak))
```
should it also be used here?

- https://github.com/cnlohr/ch32v003fun/blob/f9b5a28447533c5bf112a09a86859a7ade690464/ch32v003fun/ch32v003fun.c#L1369
- https://github.com/cnlohr/ch32v003fun/blob/f9b5a28447533c5bf112a09a86859a7ade690464/ch32v003fun/ch32v003fun.c#L1637-L1641

Does the placement of this attribute in a symbol declaration matter? This PR puts it at the start of symbol declarations/definitions - e.g.:

```
WEAK int errno;

WEAK int printf( const char* format, ... )
{
	va_list args;
	va_start( args, format );
	@@ -110,12 +112,12 @@ int printf( const char* format, ... )
	return ret_status;
}
````

whereas existing use puts it at the end, e.g.:

- https://github.com/cnlohr/ch32v003fun/blob/f9b5a28447533c5bf112a09a86859a7ade690464/ch32v003fun/ch32v003fun.c#L749-L750

Wherever it is put it probably should be consistent across the `ch32v003fun.c` file and any other related sources?

niansa:
Clang tells me putting it at the end of declarations is bad and I should just put it at the start of definitons.
Apparently it's because specifically stuff like `weak` should **only** apply to the specific definition, not to all definitions as adding it to a declaration would imply.

niansa:
I would change other appearances of attributes in another PR if desired.

TommyMurphyTM1234:
> Clang tells me putting it at the end of declarations is bad and I should just put it at the start of definitons. Apparently it's because specifically stuff like `weak` should **only** apply to the specific definition, not to all definitions as adding it to a declaration would imply.

Thanks for the clarification @niansa. I haven't used LLVM/Clang much, mainly GCC, so wasn't aware of this. 👍 

> I would change other appearances of attributes in another PR if desired.

Good point. Fair enough. 🙂

cnlohr:
@niansa do you plan to update based on comments?

TommyMurphyTM1234:
> @niansa do you plan to update based on comments?

FWIW @niansa has addressed my comments and none of them ultimately require any changes to this specific PR as far as I can see.

niansa:
Sorry for the late reply. Saw this just now. GitHub notifications are kind of bugged for me.

> FWIW @niansa has addressed my comments and none of them ultimately require any changes to this specific PR

Yes.

TommyMurphyTM1234:
@niansa (hopefully you get notified on this! 😄), maybe you could close https://github.com/cnlohr/ch32v003fun/issues/374 now that this PR has been merged by @cnlohr? Thanks. 👍

PR #384, opened 2024-09-01
Ch32v30x hsusb

cnlohr:


cnlohr:
@AlexanderMandera would you like to review this, or do you just wanna gooooooooooooo?

PR #394, opened 2024-09-17
Add setjmp/longjmp support

BogdanTheGeek:
`setjmp/longjmp` implementation for `rv32e` and `rv32ima`.
The standard `setjmp.h` that came with my toolchain defines `jmp_buf` with enough room for floating point registers as well. I did not add those registers because, as far as I can tell, support for floating point units is in the very early stages in the build tools.

I'm open to advice on how we could avoid including `setjmp.h` in `ch32v003fun.h`. That would allow us to define `jmp_buf` to not contain the space required for the floating point regs (saves 12*8 bytes of data).

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -1321,6 +1321,18 @@ void handle_reset( void )
 
 #endif
 
+#if defined( __riscv_d )
+#define FLOAD( src, offset, dst ) \
+"	fld " #src ", " #offset "*8(" #dst ")\n"
+#define FSTORE( dst, offset, src ) \
+"	fsd " #dst ", " #offset "*8(" #src ")\n"
+#else
+#define FLOAD( src, offset, dst ) \
+"	flw " #src ", " #offset "*4(" #dst ")\n"
+#define FSTORE( dst, offset, src ) \
+"	fsw " #dst ", " #offset "*4(" #src ")\n"
+#endif
+
 __attribute__ ((naked)) int setjmp(jmp_buf env)
 {
 	asm volatile(
@@ -1331,7 +1343,7 @@ __attribute__ ((naked)) int setjmp(jmp_buf env)
 "	sw sp, 3*4(a0)\n"
 
     // RV32I only registers
-#if !defined( CH32V003 )
+#if !defined( __riscv_e )
 "	sw s2, 4*4(a0)\n"
 "	sw s3, 5*4(a0)\n"
 "	sw s4, 6*4(a0)\n"
@@ -1343,6 +1355,21 @@ __attribute__ ((naked)) int setjmp(jmp_buf env)
 "	sw s10, 12*4(a0)\n"
 "	sw s11, 13*4(a0)\n"
 #endif
+
+	// FPU registers
+#if defined( __riscv_f )
+	FSTORE(fs2, 14, a0)
+	FSTORE(fs3, 15, a0)
+	FSTORE(fs4, 16, a0)
+	FSTORE(fs5, 17, a0)
+	FSTORE(fs6, 18, a0)
+	FSTORE(fs7, 19, a0)
+	FSTORE(fs8, 20, a0)
+	FSTORE(fs9, 21, a0)
+	FSTORE(fs10, 22, a0)
+	FSTORE(fs11, 23, a0)
+#endif
+
 "	li a0, 0\n"
 "	ret\n"
 	);
@@ -1358,7 +1385,7 @@ __attribute__ ((naked)) void longjmp(jmp_buf env, int val)
 "	lw sp, 3*4(a0)\n"
 
     // RV32I only registers
-#if !defined( CH32V003 )
+#if !defined( __riscv_e )
 "	lw s2, 4*4(a0)\n"
 "	lw s3, 5*4(a0)\n"
 "	lw s4, 6*4(a0)\n"
@@ -1369,6 +1396,20 @@ __attribute__ ((naked)) void longjmp(jmp_buf env, int val)
 "	lw s9, 11*4(a0)\n"
 "	lw s10, 12*4(a0)\n"
 "	lw s11, 13*4(a0)\n"
+#endif
+
+	// FPU registers
+#if defined( __riscv_f )
+	FLOAD(fs2, 14, a0)
+	FLOAD(fs3, 15, a0)
+	FLOAD(fs4, 16, a0)
+	FLOAD(fs5, 17, a0)
+	FLOAD(fs6, 18, a0)
+	FLOAD(fs7, 19, a0)
+	FLOAD(fs8, 20, a0)
+	FLOAD(fs9, 21, a0)
+	FLOAD(fs10, 22, a0)
+	FLOAD(fs11, 23, a0)
 #endif
 
 "	seqz a0, a1\n" // a0 = (a1 == 0) ? 1 : 0
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -1321,6 +1321,63 @@ void handle_reset( void )
 
 #endif
 
+int setjmp(jmp_buf env)
+{
+	asm volatile(
+	// Common registers
+"	sw ra, 0*4(a0)\n"
+"	sw s0, 1*4(a0)\n"
+"	sw s1, 2*4(a0)\n"
+"	sw sp, 3*4(a0)\n"
+
+    // RV32I only registers
+#if !defined( CH32V003 )
+"	sw s2, 4*4(a0)\n"
+"	sw s3, 5*4(a0)\n"
+"	sw s4, 6*4(a0)\n"
+"	sw s5, 7*4(a0)\n"
+"	sw s6, 8*4(a0)\n"
+"	sw s7, 9*4(a0)\n"
+"	sw s8, 10*4(a0)\n"
+"	sw s9, 11*4(a0)\n"
+"	sw s10, 12*4(a0)\n"
+"	sw s11, 13*4(a0)\n"
+#endif
+	);
+
+	return 0;
+}
+
+void longjmp(jmp_buf env, int val)
+{
+    asm volatile(
+    // Common registers
+"	lw ra, 0*4(a0)\n"
+"	lw s0, 1*4(a0)\n"
+"	lw s1, 2*4(a0)\n"
+"	lw sp, 3*4(a0)\n"
+
+    // RV32I only registers
+#if !defined( CH32V003 )
+"	lw s2, 4*4(a0)\n"
+"	lw s3, 5*4(a0)\n"
+"	lw s4, 6*4(a0)\n"
+"	lw s5, 7*4(a0)\n"
+"	lw s6, 8*4(a0)\n"
+"	lw s7, 9*4(a0)\n"
+"	lw s8, 10*4(a0)\n"
+"	lw s9, 11*4(a0)\n"
+"	lw s10, 12*4(a0)\n"
+"	lw s11, 13*4(a0)\n"
+#endif
+
+"	seqz a0, a1\n" // a0 = (a1 == 0) ? 1 : 0
+"	add a0, a0, a1\n"
+"	ret\n"
+	);
+	__builtin_unreachable(); // Disable warning about no return.
+}
+
 #if defined( FUNCONF_USE_UARTPRINTF ) && FUNCONF_USE_UARTPRINTF
 void SetupUART( int uartBRR )
 {
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -1321,19 +1321,20 @@ void handle_reset( void )
 
 #endif
 
-#if defined( __riscv_d )
+#if defined( __riscv_float_abi_double )
 #define FLOAD( src, offset, dst ) \
 "	fld " #src ", " #offset "*8(" #dst ")\n"
 #define FSTORE( dst, offset, src ) \
 "	fsd " #dst ", " #offset "*8(" #src ")\n"
-#else
+#elif defined( __riscv_float_abi_single )
 #define FLOAD( src, offset, dst ) \
 "	flw " #src ", " #offset "*4(" #dst ")\n"
 #define FSTORE( dst, offset, src ) \
 "	fsw " #dst ", " #offset "*4(" #src ")\n"
+#else // Soft float
 #endif
 
-__attribute__ ((naked)) int setjmp(jmp_buf env)
+__attribute__ ((naked)) int setjmp( jmp_buf env )
 {
 	asm volatile(
 	// Common registers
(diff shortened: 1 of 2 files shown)


BogdanTheGeek:
For now, the code doesn't with with `-Os` because the compiler inlines it so `ra` doesn't point where it should.
I think this would be easier to implement in a separate assembly file.

Example program(works with -O0):
```c
#include "ch32v003fun.h"
#include <stdio.h>
#include <setjmp.h>
static jmp_buf buf = { 0 };

void func()
{
	printf( "Welcome to GeeksforGeeks\n" );

	// Jump to the point setup by setjmp
	longjmp( buf, 1 );

	printf( "Geek2\n" );
}

int main()
{
	SystemInit();
	WaitForDebuggerToAttach();

	printf( "Start\n" );

	// Setup jump position using buf and return 0
	if ( setjmp( buf ) )
	{
		printf( "Geek3\n" );
	}
	else
	{
		printf( "Geek4\n" );
		func();
	}
	printf( "End\n" );
	while ( 1 )
	{
		;
	}
	return 0;
}
```

BogdanTheGeek:
All working now.

TommyMurphyTM1234:
Just curious...

> `setjmp/longjmp` implementation for `rv32e` and `rv32ima`. The standard `setjmp.h` that came with my toolchain

What toolchain?

> defines `jmp_buf` with enough room for floating point registers as well.

I wonder why that is when neither architecture string, `rv32e` or `rv32ima`, specifies a floating point extension (`f` or `d` (or `q`), or even `Zfinx`). What `-mabi=...` ABI string are you using?

> I did not add those registers because, as far as I can tell, support for floating point units is in the very early stages in the build tools.

I'm curious as to how you concluded this? RISC-V floating point extensions were ratified a long time ago and support for them should be stable/complete in any recent toolchain version.

BogdanTheGeek:
@TommyMurphyTM1234 
> What toolchain?
`riscv-gnu-toolchain`
regarding floating point, here is part of the `<machine/setjmp.h>` that comes with my toolchain:
```c
#ifdef __riscv
/* _JBTYPE using long long to make sure the alignment is align to 8 byte,
   otherwise in rv32imafd, store/restore FPR may mis-align.  */
#define _JBTYPE long long
#ifdef __riscv_32e
#define _JBLEN ((4*sizeof(long))/sizeof(long))
#else
#define _JBLEN ((14*sizeof(long) + 12*sizeof(double))/sizeof(long))
#endif
#endif
```
> I'm curious as to how you concluded this?

`ch32v003fun.mk` doesn't setup separate compilation flags for floating point support. The only one I can see is `-DDISABLE_FLOAT` which is not defined for some targets. The makefile also doesn't differentiate between `f/d/q` extensions and I am not familiar with all of WCH's lineup to figure out which should be used.

ch32v003fun/ch32v003fun.h
```diff
@@ -13794,6 +13794,10 @@ void SystemInit(void);
 // Put an output debug UART on Pin D5.
 // You can write to this with printf(...) or puts(...)
 
+#include <setjmp.h>
```
cnlohr, line 13797: Why do you #include this here?  Don't you only need the forward definitions?  (I'd rather not blindly #include extra headers.

BogdanTheGeek: Why repeat the typedef logic?, i could inlcude <machine/setjmp.h> which just contains the definition for jmp_buf.

BogdanTheGeek:
@TommyMurphyTM1234 
> Yes - but what version and where did you get it?

```
$ riscv64-unknown-elf-gcc --version
riscv64-unknown-elf-gcc (gc891d8dc2) 13.2.0
```

> I still don't understand. If the architecture used at compile time is -march=rv32e (or that architecture is the toolchain's default) then that should cause the __riscv_32e macro to be defined so the first definition of _JBLEN, without space for the FP registers, should be used.

Yes, I can switch to checking for `__riscv_32e`, but gcc still doesn't differentiate between `rv32ima` and `rv32imaf`, so they would all have to allocate the extra 12 double floating point regs in the `jmp_buf`. I guess I could try to commit to upstream, but barely anyone will have that version, so maybe it does make sense to just get rid of the `<setjmp.h> and just define the type in `ch32v003fun.h` ?

BogdanTheGeek:
```
>>> riscv64-unknown-elf-gcc -dM -E -march=rv32e -mabi=ilp32e - < /dev/null | grep -i RISCV
#define __riscv 1
#define __riscv_float_abi_soft 1
#define __riscv_cmodel_medlow 1
#define __riscv_32e 1
#define __riscv_xlen 32
#define __riscv_e 1009000
#define __riscv_abi_rve 1
#define __riscv_arch_test 1
>>> riscv64-unknown-elf-gcc -dM -E -march=rv32ima -mabi=ilp32 - < /dev/null | grep -i RISCV
#define __riscv 1
#define __riscv_atomic 1
#define __riscv_float_abi_soft 1
#define __riscv_cmodel_medlow 1
#define __riscv_mul 1
#define __riscv_muldiv 1
#define __riscv_xlen 32
#define __riscv_m 2000000
#define __riscv_a 2001000
#define __riscv_i 2001000
#define __riscv_arch_test 1
#define __riscv_div 1
>>> riscv64-unknown-elf-gcc -dM -E -march=rv32imafc -mabi=ilp32f - < /dev/null | grep -i RISCV
#define __riscv 1
#define __riscv_atomic 1
#define __riscv_float_abi_single 1
#define __riscv_cmodel_medlow 1
#define __riscv_mul 1
#define __riscv_muldiv 1
#define __riscv_xlen 32
#define __riscv_fsqrt 1
#define __riscv_m 2000000
#define __riscv_fdiv 1
#define __riscv_a 2001000
#define __riscv_c 2000000
#define __riscv_f 2002000
#define __riscv_i 2001000
#define __riscv_zicsr 2000000
#define __riscv_compressed 1
#define __riscv_flen 32
#define __riscv_arch_test 1
#define __riscv_div 1
```

ok, looks like I might be able to figure out capabilities based on march and mabi.

TommyMurphyTM1234:
> Yes, I can switch to checking for `__riscv_32e`, but gcc still doesn't differentiate between `rv32ima` and `rv32imaf`, so they would all have to allocate the extra 12 double floating point regs in the `jmp_buf`. 

Ah - OK- I think I get it now.
The check here should really be to only allocate space for the FPU registers if the target architecture specifies that an FP extension is supported?
```
...
#ifdef __riscv_32e
#define _JBLEN ((4*sizeof(long))/sizeof(long))
#else
#define _JBLEN ((14*sizeof(long) + 12*sizeof(double))/sizeof(long))
#endif
...
```
But I'm not sure if that means something like (pseudocode):
```
#if <target supports FP>
#define _JBLEN ((14*sizeof(long) + 12*sizeof(double))/sizeof(long))
#elif <target is __riscv_32e>
#define _JBLEN ((4*sizeof(long))/sizeof(long))
#elif <any other (non FPU enabled) target>
#define _JBLEN ((14*sizeof(long)
#endif
```
Or some variant on this?
I might raise it in the `riscv-gnu-toolchain` repo to get the opinion of the developers there even if such changes would be done in upstream GCC these days.

> I guess I could try to commit to upstream

Ultimately that would be a good idea even if a temorary/local fix is done here in `ch32v003fun`.

BogdanTheGeek:
@TommyMurphyTM1234 what about this ^

TommyMurphyTM1234:
FWIW I've opened an issue here:

- https://github.com/riscv-collab/riscv-gnu-toolchain/issues/1555

@BogdanTheGeek - I'll try to review your latest commit later. Thanks.

TommyMurphyTM1234:
Ok, looks like modifying `setjmp.h` upstream isn't really a suitable solution after all: 

- https://github.com/riscv-collab/riscv-gnu-toolchain/issues/1555#issuecomment-2367084487

So, a local change to how this works for `ch32v003fun` is probably the best option after all. 

I still haven't had a chance to review your latest commit @BogdanTheGeek but maybe the comments in the linked `riscv-gnu-toolchain` issue may alter/influence what's needed? In particular, the fact that the ABI and not the architecture seems to be what's relevant when sizing `_JBLEN`?

BogdanTheGeek:
@TommyMurphyTM1234 thanks for reaching out to upstream. I think my solution is better for this application. I have tested it agains all the targets and it produces the asm I expect based on arch and float support.

ch32v003fun/ch32v003fun.h
```diff
@@ -13794,7 +13794,29 @@ void SystemInit(void);
 // Put an output debug UART on Pin D5.
 // You can write to this with printf(...) or puts(...)
 
-#include <setjmp.h>
+#ifdef __riscv
```
TommyMurphyTM1234, line 13797: Is there any point in this check since `ch32v003fun.h` is implicitly for RISC-V?

ch32v003fun/ch32v003fun.h
```diff
@@ -13794,7 +13794,29 @@ void SystemInit(void);
...
+#ifdef __riscv_e
+#define _JBLEN ((4*sizeof(long))/sizeof(long))
+#elif defined(__riscv_d)
+#define _JBLEN ((14*sizeof(long) + 12*sizeof(double))/sizeof(long))
+#elif defined(__riscv_f)
+#define _JBLEN ((14*sizeof(long) + 12*sizeof(float))/sizeof(long))
+#else
+#define _JBLEN ((14*sizeof(long))/sizeof(long))
+#endif
+#endif
```
TommyMurphyTM1234, line 13810: If this works then ir's probably fine but based on the comments here (https://github.com/riscv-collab/riscv-gnu-toolchain/issues/1555) the relevant checks should probably be done based on the ABI macro rather that other macros?

TommyMurphyTM1234 (commented):
Hi @BogdanTheGeek - some comments on the latest commit in case they help at all. :-)

ch32v003fun/ch32v003fun.c
```diff
@@ -1321,6 +1321,18 @@ void handle_reset( void )
...
+"	fld " #src ", " #offset "*8(" #dst ")\n"
+#define FSTORE( dst, offset, src ) \
+"	fsd " #dst ", " #offset "*8(" #src ")\n"
+#else
+#define FLOAD( src, offset, dst ) \
+"	flw " #src ", " #offset "*4(" #dst ")\n"
+#define FSTORE( dst, offset, src ) \
+"	fsw " #dst ", " #offset "*4(" #src ")\n"
+#endif
+
```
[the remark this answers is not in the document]
BogdanTheGeek: if you enable Double support, the Float support is also enabled (aka, F is always defined if D is defined), so bellow I only need to check if F is defined. I wrote it like this to avoid having to put in every single `elif` in the main sections. If someone needs to add Quad Float support, they only need to change it here, the code bellow remains valid and correct.

ch32v003fun/ch32v003fun.c
```diff
@@ -1343,6 +1355,21 @@ __attribute__ ((naked)) int setjmp(jmp_buf env)
...
+	FSTORE(fs3, 15, a0)
+	FSTORE(fs4, 16, a0)
+	FSTORE(fs5, 17, a0)
+	FSTORE(fs6, 18, a0)
+	FSTORE(fs7, 19, a0)
+	FSTORE(fs8, 20, a0)
+	FSTORE(fs9, 21, a0)
+	FSTORE(fs10, 22, a0)
+	FSTORE(fs11, 23, a0)
+#endif
```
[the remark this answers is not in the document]
BogdanTheGeek: See the other comment.

ch32v003fun/ch32v003fun.h
BogdanTheGeek: Just habit, I can remove it if you want.

ch32v003fun/ch32v003fun.c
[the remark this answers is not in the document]
TommyMurphyTM1234: > if you enable Double support, the Float support is also enabled (aka, F is always defined if D is defined),

I didn't think that this was necessarily the case - I thought that F and D could be implemented separately or together?

ch32v003fun/ch32v003fun.h
BogdanTheGeek: Would you rather me use these?
```c
#define __riscv_float_abi_double 1 // ilp32d
#define __riscv_float_abi_single 1 // ilp32f
#define __riscv_float_abi_soft 1   // ilp32
#define __riscv_abi_rve 1          // ilp32e
```
Would be nice if GCC actually wrote down their define logic...

BogdanTheGeek:
@TommyMurphyTM1234 ping me a message on the discord server if you wanna chat more about this, it's easier to brainstorm.

ch32v003fun/ch32v003fun.c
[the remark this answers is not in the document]
TommyMurphyTM1234: > I didn't think that this was necessarily the case - I thought that F and D could be implemented separately or together?

Sorry, I was wrong: 

- https://github.com/riscv/riscv-isa-manual/blob/main/src/naming.adoc

> Some ISA extensions depend on the presence of other extensions, e.g., "D" depends on "F" and "F" depends on "Zicsr".

ch32v003fun/ch32v003fun.h
TommyMurphyTM1234: > Would you rather me use these?
> 
> ```c
> #define __riscv_float_abi_double 1 // ilp32d
> #define __riscv_float_abi_single 1 // ilp32f
> #define __riscv_float_abi_soft 1   // ilp32
> #define __riscv_abi_rve 1          // ilp32e
> ```
> 
> Would be nice if GCC actually wrote down their define logic...

I presume you're referring to these? 

- https://github.com/riscv-non-isa/riscv-c-api-doc/blob/main/src/c-api.adoc#abi-related-preprocessor-definitions

TommyMurphyTM1234:
> @TommyMurphyTM1234 ping me a message on the discord server if you wanna chat more about this, it's easier to brainstorm.

Thanks, but I don't use Discord - sorry:

- https://blog.mattcen.com/2023/07/04/stop-using-discord-for-your-open-source-communities/

ch32v003fun/ch32v003fun.h
BogdanTheGeek: thank you!

BogdanTheGeek:
@TommyMurphyTM1234 Anything else needed for this to be merged?

TommyMurphyTM1234:
> @TommyMurphyTM1234 Anything else needed for this to be merged?

Hi @BogdanTheGeek - not that I can see. Looks good to me now. Thanks a lot for accepting my review feedback. 👍

cnlohr:
Looks great! No comments!

cnlohr:
Sorry, I spoke too soon @BogdanTheGeek  It breaks the OSX/Linux minichlink builds.

You will need to protect your longjmp/setjmp with the following.

```
#if defined(__riscv) || defined(__riscv__) || defined( CH32V003FUN_BASE )
```

As to not break the mac/linux build.

Alternatively you could put the definition around here: https://github.com/cnlohr/ch32v003fun/blob/master/ch32v003fun/ch32v003fun.h#L13525

Since it really is an architecture helper.

PR #401, opened 2024-09-25
Alias the EXTEND register to match reference manual

cnlohr:


ch32v003fun/ch32v003fun.h
```diff
@@ -1222,6 +1222,13 @@ typedef struct
     __IO uint32_t EXTEN_CTR;
 } EXTEN_TypeDef;
 
+/* The reference manual for the 32v2xx/v3xx refernce this as "CTR" on "EXTEND" so adding an alias here. for any processor. */
```
TommyMurphyTM1234, line 1225: Typos:
- "refernce" should be ”reference"
- "on" should be "or"?
- "32v2xx" should be "ch32v2xx"?
- ". for any processor" doesn't seem to read properly?

ch32v003fun/ch32v003fun.h
```diff
@@ -2757,6 +2764,7 @@ typedef struct
 #define OB                                      ((OB_TypeDef *)OB_BASE)
 #define ESIG                                    ((ESG_TypeDef *)ESIG_BASE)
 #define EXTEN                                   ((EXTEN_TypeDef *)EXTEN_BASE)
+#define EXTEND                                  ((EXTEND_TypeDef *)EXTEN_BASE)  // Alias to EXTEN
```
TommyMurphyTM1234, line 2767: Typo
- `EXTEN` should be `EXTEND`

TommyMurphyTM1234 (commented):
Some small comments which I hope are useful.

ch32v003fun/ch32v003fun.h
cnlohr: No, this is what I intended to do.  I am aliasing `EXTEND` to `EXTEN`, but with different register names.

ch32v003fun/ch32v003fun.h
cnlohr: Fixed in new rev.  Thanks for the feedback.  I will wait for it to pass CI.

PR #405, opened 2024-10-03
Split linker script

monte-monte:
Making this PR to show what exactly I meant. Feel free to discard it.


TommyMurphyTM1234:
> Making this PR to show what exactly I meant. 

What you meant where? What's the background/context to this PR? Thanks.

monte-monte:
> > Making this PR to show what exactly I meant.
> 
> What you meant where? What's the background/context to this PR? Thanks.

Sorry, it was meant as continuation of a conversation in discord. The Idea is to separate memory portion of ld script in a separate file and include common part from it. This way we can have custom ld scripts if needed and include the common portion. It is a controversial proposal, so I just wanted to clarify in code how it should work.

cnlohr:
To clarify the way I would like to keep it is for people to just use custom LD scripts like in the bootloader example.

PR #410, opened 2024-10-10
Fix detection of projects to build

maxgerhardt:
Drive by fix

Restricts search of folders with `Makefile` file in them to only 2 folders deep as seen from the root, so `examples/blink/Makefile` is fine but ont `examples/blink/nested_util/Makefile`


cnlohr:
Thank you for doing this!

PR #422, opened 2024-10-19
Fix USB bootloader programmer

monte-monte:
``if( is_flash && memcmp( &eps->commandbuffer[60], blob, 1 ) ) goto verifyfail;`` I believe this should compare with respbuffer, otherwise it makes no sense.

``usleep(1000);`` before writing feature report so bootloader can process previous commands and minichlink doesn't throw error about retrying.

``if (eps->no_get_report) return r;`` so we don't wait to get feature report that will never come because bootloader finished working and usb connection throws ``ioctl (GFEATURE): No such device`` errors at us until timeout.


cnlohr:
Looks good to me.

PR #427, opened 2024-10-23
fix the adc code without debug printf

eeucalyptus:


cnlohr:
Oops! Sorry about that.  Thank you for fixing this.

PR #430, opened 2024-10-29
Add timeout to `WaitForDebuggerToAttach`

BogdanTheGeek:
Added timeout in milliseconds argument and a return value to indicate weather timeout was reached.

Possible future enhancements:
 - disable debug `printf` if timeout reached by default.

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -1578,9 +1578,17 @@ void SetupDebugPrintf()
 	*DMDATA0 = 0x80;
 }
 
-void WaitForDebuggerToAttach()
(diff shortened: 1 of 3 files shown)


eeucalyptus:
I'm not exactly happy with the inline defines for the type. 
Would 

```
#if defined(CH32V20x) || defined(CH32V30x)
uint64_t start = SysTick->CNT;
#elif defined(CH32V10x) || defined(CH32X03x)
uint32_t = SysTick->CNTL;
#else
uint32_t start = SysTick->CNT
#endif
```

be fine too? If I'm being too pedantic lmk, I'd just accept this PR then xD

eeucalyptus:
Yes I am being too pedantic ;)

PR #431, opened 2024-10-29
Fix C functions that take any arguments

BogdanTheGeek:
From C11 onwards:
> The use of function declarators with empty parentheses (not prototype-format parameter type declarators) is an obsolescent feature.

```c
void foo(); // can be called with any arguments
void bar( void ); // can only be called with no arguments
```


cnlohr:
This was such an annoying and ugly change to the standard. I hate it so much.

But I am not in a place to protest :-/

PR #433, opened 2024-11-03
Double-initialize shadow regs at boot

cnlohr:


cnlohr:
I tested this in a couple scenarios, and it really seems to be best.

PR #435, opened 2024-11-06
Cleanup (macro algebra) the 30x interrupts.

cnlohr:
Mind double-checking me @hexaedron ?


hexaedron:
@cnlohr I don't have any 30x, so i can't check if that works on hardware. But that change seems correct for me. ✅

PR #436, opened 2024-11-06
Removed excessive define logic from interrupt handlers block

hexaedron:
As we discussed in Discord.


cnlohr:
This looks great! Thanks!

PR #438, opened 2024-11-06
Fix HIDAPI on macOS

BogdanTheGeek:
I was not able to detect the usb bootloader using the implementation of hidapi bundled with the project on macOS 14.6.1 on M1.
I was able to enumerate the device, but hid_open() would return a null pointer.
Long story short, the version of `hidapi.c` is too old to work on macOS, but linking with hidapi provided by homebew works.


BogdanTheGeek:
bad method, trying something else

PR #439, opened 2024-11-06
Try updating to new hidapi

cnlohr:


BogdanTheGeek:
Tested on macOS 14.6.1 on M1 and I can confirm it fixes the issue with the usb bootloader and works as expected.

PR #449, opened 2024-11-16
Fix bad GPIO_pinMode overview in branchless GPIO

niansa:


cnlohr:
I still don't quite understand how GPIO_pinMode is used here... @recallmenot do you mind double checking this?

PR #450, opened 2024-11-20
Add HPE support

hexaedron:
As discussed with Max and Cnlohr in Discord.


cnlohr:
If anyone else sees this and is curious the way this is used is here: https://github.com/cnlohr/ch32v003fun/blob/4e890dd053cfff8c3ba4b81fc182a7d4e69805aa/ch32v003fun/ch32v003fun.c#L1012-L1019

PR #455, opened 2024-11-26
Fix unitialized WCH-LinkE errors

monte-monte:
If LE read or write command fails check internal state of programmer and chip id to determine if programmer was initialized in this run. If not, send magic command in hope to fix things. Works when you just plugged WCH-LinkE into the USB and running minichlink with command that skips startup.

### DIFF
--- a/minichlink/pgm-wch-linke.c
+++ b/minichlink/pgm-wch-linke.c
@@ -238,21 +238,24 @@ int LEWriteReg32( void * dev, uint8_t reg_7_bit, uint32_t command )
 	wch_link_command( devh, req, sizeof(req), &resplen, resp, sizeof(resp) );
 	if( resplen != 9 || resp[8] == 0x02 || resp[8] == 0x03 ) //|| resp[3] != reg_7_bit )
 	{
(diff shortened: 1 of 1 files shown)


minichlink/pgm-wch-linke.c
```diff
@@ -264,14 +274,24 @@ int LEReadReg32( void * dev, uint8_t reg_7_bit, uint32_t * commandresp )
 	*commandresp = ( rbuff[4]<<24 ) | (rbuff[5]<<16) | (rbuff[6]<<8) | (rbuff[7]<<0);
 	if( transferred != 9 || rbuff[8] == 0x02 || rbuff[8] == 0x03 ) //|| rbuff[3] != reg_7_bit )
 	{
-		fprintf( stderr, "Error setting write reg. Tell cnlohr. Maybe we should allow retries here?\n" );
-		fprintf( stderr, "RR: %d :", transferred );
-		int i;
-		for( i = 0; i < transferred; i++ )
+		struct InternalState *iss = (struct InternalState *)( ( (struct ProgrammerStructBase *)dev )->internal );
+		if ( !iss->target_chip_id && !iss->statetag )
```
cnlohr, line 278: statetag doesn't mean anything here.

PR #459, opened 2024-11-30
[CH32X035] Add UART echo example with DMA and IDLE interrupts

jurriaan:
Adds an example demonstrating efficient UART communication using DMA and IDLE line detection on the CH32X035 MCU.

The example utilizes DMA for data transfer and the UART's IDLE line detection feature to identify the end of a data packet.


cnlohr:
Looks good! Do you use Discord?  If so, you should join us, as we discuss the direction of 003fun, you can ping me `@cnlohr`

PR #465, opened 2024-12-02
Add TIM1 complementary outputs example

bikeNomad:
This example shows how to set up Timer 1 (the Advanced Control Timer) to generate
Pulse-Width Modulation (PWM) with complementary outputs on two output pins.
This mode would be used, for example, to drive a motor with a H-bridge.

This program shows the effect of changing the PWM deadtime settings with complementary outputs.

### DIFF
--- a/examples/tim1_pwm_complementary_outputs/.gdbinit
+++ b/examples/tim1_pwm_complementary_outputs/.gdbinit
@@ -1,0 +1,2 @@
+file tim1_pwm.elf
+target extended-remote localhost:3333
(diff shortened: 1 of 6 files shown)


examples/tim1_pwm_complementary_outputs/.gdbinit
```diff
@@ -0,0 +1,2 @@
+file tim1_pwm.elf
+target extended-remote localhost:3333
```
cnlohr, line 2: I don't believe this file should be committed.  cc @maxgerhardt

bikeNomad: OK, deleted file.

cnlohr:
Looks good!

PR #468, opened 2024-12-04
Add time alignment to stabilize ADC sampling

cnlohr:


cnlohr:
I tested this and it made a world of difference.  This is a "worst case" of both, real - but worst. for counts from adc_cap_touch example.

image

PR #470, opened 2024-12-04
Modernize tim1_complementary_outputs example

bikeNomad:
I realized that there were now Arduino-style #defines for the PINs. This commit changes the pin initialization in my example to use them.


cnlohr:
Wonderful!

PR #472, opened 2024-12-07
Fix copy-paste error in Ardulink init

maxgerhardt:
Discovered by @monte-monte in https://github.com/cnlohr/ch32v003fun/compare/master...monte-monte:ch32v003fun:usb_terminal, looks like a clear bug.

Used when explicitly giving `./minichlink -C arduilink`


TommyMurphyTM1234:
Yes, I had previously asked about this issue here: 

- https://github.com/cnlohr/ch32v003fun/issues/391#issuecomment-2349570203

> Although, even if it should work, this looks odd - why `TryInit_B003Fun()` instead of `TryInit_Ardulink()` for the Ardulink case?

PR #477, opened 2024-12-15
minichlink: add ch641

recalci:
Tests passed. 
The address can be find in DBGMCU_GetCHIPID in \EVT\EXAM\SRC\Peripheral\src\chxxx_dbgmcu.c


cnlohr:
I can't get to this right now but can start looking into it in the next few days.

As a warning, we are going to be reorganizing a lot and splitting processors into their own headers.  It may make sense to wait until after that happens to add new processor support.

recalci:
Thanks for letting me know about the plans.
I have v003, ch641, ch643, x035, and l103.
I'd love to help with the process if there's anything I can do!

cnlohr:
Closed in favor of #503

PR #478, opened 2024-12-20
Fix missing USART constants for CH32X03x

jurriaan:
Noticed that although the USAR(T)?_BASE constants where defined, the USAR(T)? constants weren't for the CH32X03x.

Verified USART3 on my CH32X035. The rest I haven't tested yet, but seems to match the reference manual.

Oh and the 4th UART peripheral is called USART4 on CH32X035, not UART4. Might be nice to fix that as well, but for now the extra USART peripherals are at least accessible with this PR.


cnlohr:
Thanks

PR #480, opened 2024-12-25
changing standard mode of the CH32x035 to machine mode

yves5141:
According to the reference manual, ch32x035 has a PMP and is by default in user mode which prevents from reading some registers, mstatus for example.
This change set the default mode to machine mode.

### DIFF
--- a/ch32v003fun/ch32v003fun.c
+++ b/ch32v003fun/ch32v003fun.c
@@ -1023,6 +1023,13 @@ void handle_reset( void )
 	or a0, a0, a3\n\
 	csrw mtvec, a0\n" 
 	: : : "a0", "a3", "memory");
+//This changes the standard mode of the CH32x035 to machine mode
+#if defined ( CH32X03x )
+asm volatile(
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
Presumably relates to this issue? 

- https://github.com/cnlohr/ch32v003fun/issues/479#issuecomment-2561740505

TommyMurphyTM1234:
Please see my comment here as to why I'm not convinced by this change. 

- https://github.com/cnlohr/ch32v003fun/issues/479#issuecomment-2561878706

cnlohr:
We set mstatus here. 

https://github.com/cnlohr/ch32v003fun/blob/master/ch32v003fun/ch32v003fun.c#L1012-L1020

Why would you set it as well later?

cnlohr:
Also are you sure you want to be setting MPP and XS?  I don't think you do.  That sounds like a holdover from some completely unrelated processor.

cnlohr:
If this is still an issue we will need to open a new PR, since the trees have diverged.

PR #485, opened 2025-01-01
Enable `Zicsr` in `__get_MEPC()`

TommyMurphyTM1234:
Fix `__get_MEPC()` so that, like other CSR get/set helper functions, it enables `Zicsr` extension for GCC > 10.

Fixes:

- https://github.com/cnlohr/ch32v003fun/issues/483

### DIFF
--- a/ch32v003fun/ch32v003fun.h
+++ b/ch32v003fun/ch32v003fun.h
@@ -13513,10 +13513,13 @@ static inline void __set_MSCRATCH(uint32_t value)
  */
 static inline uint32_t __get_MEPC(void)
 {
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
I suspect that this commit omitted to apply the same change to `__get_MEPC` that was applied to other CSR get/set helper functions?

- https://github.com/cnlohr/ch32v003fun/commit/4b835d34b4dfcd9c02f3a6a747136d6b316688e7

TommyMurphyTM1234:
Thanks for accepting/merging this PR @cnlohr. 👍

PR #492, opened 2025-01-14
Udev rules and related documentation

imbus64:
From my understanding, udev should simply match the first available group, so specifying duplicates like this should not be a problem. However, a sed/envsubst solution might be better from a maintainer point of view. Tested and working on F41 (dialout), but needs a check on a plugdev system.


cnlohr:
This seems good to me!

PR #496, opened 2025-01-18
CH32V003: Add accessors for DBGMCU_CR CSR register and correct its bit definitions.

bikeNomad:
This PR adds `__get_DEBUG_CR()` and `__set_DEBUG_CR()` functions to access the `DBGMCU_CR` CSR register in the CH32V003.
It also corrects the bit definitions for that register per the Reference Manual section 18.2.


cnlohr:
Thanks, this is great!

PR #499, opened 2025-01-20
Add example of using DMA for phase-shifted PWM outputs.

bikeNomad:
This example shows how to use DMA with TIM1 to generate phase-shifted PWM outputs. This would be used with a full-bridge driver to generate controllable differential outputs over a wide range of duty cycles (vs. complementary PWM, which only produces up to 50% differential duty cycle)


cnlohr:
Thanks for doing this, this looks great.

PR #500, opened 2025-01-21
replaced direct pin call with gpio variable

davedarko:
There was a stray pin call that wasn't using the gpio_pin variable in the dma_gpio_ws2812b example for the CH32V003.


cnlohr:
Thanks for catching this.

PR #501, opened 2025-01-23
minichlink: Active polling of Ardulink device

Baldanos:
Hi !

This PR allows to actively poll an Ardulink device instead of waiting for the '!' character.
The poll is done using the '?' command, which returns a '+' character.

Submitting this because the tool I develop uses a bus pirate-like interface. Switching to ardulink mode requires the user quit the terminal to run minichlink, and the initial '!' usually gets lost.

I also thought about changing the serial port behavior in minichlink (using timeouts on serial reads), but this solution is simplier and works on the original ardulink as well.


cnlohr:
Looks good, but just FYI there won't be any minichlink compiles for a bit.

PR #505, opened 2025-02-01
Fix return value of strcpy() and strncpy(): they should return a poin…

pa3fwm:
…ter to the beginning of the destination string, not the end.


cnlohr:
Oops. Thanks.

PR #506, opened 2025-02-03
Convert ch32v003fun -> ch32fun

cnlohr:
* Split apart header into specific hardware devices.


cnlohr:
I still need to make sure this all works, but, I'd like some more eyeballs on this.

eeucalyptus:
I'll take a close look tonight (UTC+1)

recalci:
I suggest considering splitting headers by peripherals, same as WCH’s official approach. We can use ch32xxxhw.h to manage which peripherals are available for specific model. This might make merging, updating, and checking easier.

cnlohr:
> I suggest considering splitting headers by peripherals, same as WCH’s official approach. We can use ch32xxxhw.h to manage which peripherals are available for specific model. This might make merging, updating, and checking easier.

No.  This would make it way too voluminous.  

Additionally, I want these files to be something people can use as a quick reference to see what features a given processor has instead of having to trull through lots of headers.

cnlohr:
Do we want to include changes to ch32fun.c, reorganizing how we setup the default ISR + startup code is organized.

(+ changing the printf-by-default code in the hard-fault-handler)

hexaedron:
What do you mean by reorganizing?

cnlohr:
Right now, if you look at the vector table, and .start code, it is difficult to tell what is used where.  

CC @dwillmore

dwillmore:
I did think it could get split out, yes.

eeucalyptus:
All projects on 003, 035 and 203 I have are working fine with the changes

cnlohr:
Do you think we should make more deep changes in ch32fun.c before committing this?  Or commit this as is since it will match the new structure and people can begin to develop against it, and we can rename the repository.

hexaedron:
IMHO we should do changes step by step. I don't see any reasons to make one giant commit instead of a series of commits.
But this is up to you.

orbitalair:
I checked my 003 project.  Built and ran ok, but its about 100bytes larger?
Anyone see this on blink or other programs?   I will try and compare blinks later.
Were any compile flags changed?

orbitalair:
Nevermind.   False Alarm.  Blink seems ok.   I will investigate my own app code, i must have changed something.

cnlohr:
I will squash and merge this this evening if there are no other concerns.

PR #510, opened 2025-02-06
Split vectors out by processor

cnlohr:


cnlohr:
How do people feel about this?  I think it's much cleaner/clearer what's what instead of navigating through the tree of craziness.

I spot checked most of the projects to make sure they were the same.

dwillmore:
It looks pretty good to me.

hexaedron:
Looks great! Much more clean and easy to understand than before.

PR #512, opened 2025-02-06
Add missing defines for EXTI on CH32X03x

jurriaan:
Including a very simple example.

Somehow the reference does not match reality for the CH32X035 at least, as to select PC for an external interrupt the bits are b10, not b11 as described in the reference manual (same for PB, that's b01 instead of b10).

image


cnlohr:
Looks great!! Thank you!

PR #523, opened 2025-02-12
OLED passthrough errors + allow custom startup code.

cnlohr:
Make it possible to customize startup code for OLED displays, and pass through errors.


cnlohr:
Code doesn't work with my target OLED.  Still working on it.

cnlohr:
Thank you for testing, @emeb

PR #533, opened 2025-02-23
Move the gp register, to allow the compiler to leverage gp to access …

cnlohr:
…the first part of ram.  This produces smaller executables in general.  I don't know why the original startup code pointed at 0x20000800.  You can reach all of the first part of ram by setting it to 0x200003fc

### DIFF
--- a/ch32fun/ch32fun.ld
+++ b/ch32fun/ch32fun.ld
@@ -173,11 +173,11 @@ SECTIONS
     .data :
     {
       . = ALIGN(4);
+      __global_pointer$ = . + 0x3fc; /* This gets set in the startup code.  This allows -mrelax'd code to be smaller by acting as a sort of quick reference in the gp register. */
       *(.gnu.linkonce.r.*)
       *(.data .data.*)
       *(.gnu.linkonce.d.*)
       . = ALIGN(8);
-      PROVIDE( __global_pointer$ = . + 0x800 );
       *(.sdata .sdata.*)
       *(.sdata2*)
       *(.gnu.linkonce.s.*)


TommyMurphyTM1234:
> I don't know why the original startup code pointed at 0x20000800.

I don't know if this is discussion about `$gp` might help at all? 

- https://groups.google.com/a/groups.riscv.org/g/sw-dev/c/60IdaZj27dY?pli=1

monte-monte:
Yeah, it seems they just copepasted it from the chips with bigger ram. But if gp is used as described here https://gnu-mcu-eclipse.github.io/arch/riscv/programmer/#the-gp-global-pointer-register why does this change produces effect? If all 2K fit in *negative* region of 12bit offset?

cnlohr:
It looks like it might be primarily a GCC bug.  It SHOULD be able to leverage 0x20000800 But, in practice, even with GCC v14.2.0-3 it still doesn't know how to use it appropriately.

TommyMurphyTM1234:
> It looks like it might be primarily a GCC bug. It SHOULD be able to leverage 0x20000800 But, in practice, even with GCC v14.2.0-3 it still doesn't know how to use it appropriately.

Do you happen to have a reproducible test case? If this is a bug in GCC then it should ideally be logged upstream. Although isn't the `$gp` primarily used for linker relaxation so any issue in this context may actually be with the Binutils linker rather than GCC per se?

monte-monte:
@TommyMurphyTM1234 I compiled a project with different values for gp. Only when offset is a big negative number gcc seems to ignore the existence of gp. If you set gp = 0 then it works (the positive portion of 12bit)

cnlohr:
Before submitting this, I also want to try adding `-ffunction-sections` - there seems to be some situations where GCC will not strip unused functions.

Also, as we discussed on the Discord.  We cannot find a reason why GCC can't use the full +/- 2k Range, but, it does seem to effectively use the +/- 1k range.

cnlohr:
We already use -ffunction-sections it turned out, and I was just misusing the cflags in my test.  All is well.  Still need to make this change so I am submitting it.

PR #536, opened 2025-02-28
Add exemption for gcc<14 for not using zmmul extension

cnlohr:


ch32fun/ch32fun.mk
```diff
@@ -36,6 +36,8 @@ endif
 CFLAGS?=-g -Os -flto -ffunction-sections -fdata-sections -fmessage-length=0 -msmall-data-limit=8
 LDFLAGS+=-Wl,--print-memory-usage -Wl,-Map=$(TARGET).map
 
+GCCVERSIONG14 := $(shell expr `gcc -dumpversion | cut -f1 -d.` \>= 14)
```
TommyMurphyTM1234, line 39: Should this actually be `GCCVERSIONG14` or should it be `GCCVERSION14`? Why the trailing `G`?

TommyMurphyTM1234:
According to this `Zmmul` support was added to GCC starting with GCC 13.

- https://gcc.gnu.org/gcc-13/changes.html

TommyMurphyTM1234:
>```
> CFLAGS_ARCH+=-march=rv32eczmmul -mabi=ilp32e -DCH32V00x=1
> ```

Minor innocuous issue, but the RISC-V convention is usually to use `rv32ec_zmmul` rather than `rv32eczmmul`.

cnlohr:
> According to this, support for the RISC-V `Zmmul` extension was added to GCC starting with GCC 13 not 14.
> 
>     * https://gcc.gnu.org/gcc-13/changes.html

AH! Thank you! I didn't test 13.  Also I will test with rv32ec_zmmul tonight.

AND @prplz no, it still works.  It just uses rv32ec, instead.

cnlohr:
I checked, it works as far back as GCC12.

cnlohr:
I take it back, the codegen in gcc12 is jank.  gcc13 it is.

PR #544, opened 2025-03-02
Fix on windows + Allow bypass mode for ch32v006

cnlohr:


cnlohr:
Wow that was a cursed set of stuff. But now you can use the 006 with the older link-E's without a firmware upgrade.

PR #545, opened 2025-03-02
Add USB terminal functionality

monte-monte:
Here are multiple proposed improvements.
- Added ``-k`` option to skip programmer initialization. This is useful in some cases
- Added ability to specify VID/PID for use with b003 programmer option
- Now you can use command like this ``minichlink -kT -c 0x1209d003`` to launch a USB terminal and look for specific USB device
When PR to rv003usb gets merged I will update this one and also add the function to reboot into USB bootloader.


cnlohr:
Mind fixing the macosx build?  (Also still consider making it a runtime flag to allow this feature for everyone!)

monte-monte:
Yeah, fixed macos build.
This is using a runtime flag ``minichlink -kT -c 0x1209d003``. It will work as if it was b003 *programmer* but extra option ``-k`` needed to skip programmer initialization (because it's not actually a bootloader)

cnlohr:
This is a ... uuh ... lot.

cnlohr:
Have you verified normal operation with esp32-s2, including debugprintf functionality?

monte-monte:
I don't have S2. I may be missing something, but how is ``b003`` connected to esp32-s2 programmer?

cnlohr:
My concern is this
image

cnlohr:
OH I AM A BIG GOON I will accept now.

PR #546, opened 2025-03-03
Fix on windows

cnlohr:


cnlohr:
I think I want to accept this wholesale.  I just have two questions:

1. Does it make sense to have two separate install scripts? One for user and one for system?
2. Is there any way to make it so a user can "double-click" or "right-click-and" to invoke the installer?

cnlohr:
@CaiB 
I think I want to accept this wholesale.  I just have two questions:

1. Does it make sense to have two separate install scripts? One for user and one for system?
2. Is there any way to make it so a user can "double-click" or "right-click-and" to invoke the installer?

CaiB:
> Does it make sense to have two separate install scripts? One for user and one for system?

Not really, the code is pretty much the same, and being able to specify user, system, or specific path with one script seems like a better idea, and doesn't risk one getting out of date later on.

> Is there any way to make it so a user can "double-click" or "right-click-and" to invoke the installer?

Yes, this is possible as-is, assuming they want to user-install (the default). Here's what that looks like (I just ran this in a fresh VM):
xpack_user_rightclick

If they want to system-install, then they'd need to run it from a terminal to provide the launch argument.

Note that the [same caveat as Install-TCC with execution policies applies](https://github.com/cntools/Install-TCC?tab=readme-ov-file#usage), which blocks PowerShell scripts from running by default unless a user changes this setting. The right-click-to-run approach bypasses this, but running from the terminal won't.

cnlohr:
WOOT!

PR #547, opened 2025-03-03
fix ch32fun directory reference in .gitignore.

BrianPugh:
seems like this was missed in #506


cnlohr:
It was! Thank you for catching it.

PR #551, opened 2025-03-06
Add README links to BMP280 and AHT20 weather sensor drivers

pabloestrado:
I have created drivers for AHT20 and BMP280 sensors and tested it on one of my projects and would like to share it with community

https://github.com/pabloestrado/ch32fun-aht20-driver
https://github.com/pabloestrado/ch32fun-bmp280-driver


cnlohr:
thanks for adding them.  One bit of feedback for you is that people really like it when there are photos + schematics of the things in your readme. You might want to consider adding that.

PR #552, opened 2025-03-08
Fixed a typo in the debugprintfdemo for ch32v30x.

DeadBugEngineering:
Fixed a typo in the debugprintfdemo for ch32v30x.


cnlohr:
we have never really formalized the intent here.  Realistically moving forward, you shouldn't need to #define the processor in `funconfig.h` do you need to on this example?

cnlohr:
To clarify, do you mind seeing if you can still use the demo with the #define commented out, @DeadBugEngineering

cnlohr:
I guess even through we've never formalized it, there's no reason not to accept this if it makes your life more convenient.

PR #553, opened 2025-03-08
Detecting ANSI compatibility

monte-monte:
Added check for ANSI escape support. This should resolve possible edge cases.
For windows I am setting console mode to enable ANSI and if it fails, fall back to simple terminal.
https://learn.microsoft.com/en-us/windows/console/getconsolemode
Tested on win11 and win7.  Would appreciate if you test it on win10 or 8 also.

### DIFF
--- a/ch32fun/README.md
+++ b/ch32fun/README.md
@@ -33,7 +33,7 @@
 |USBPD         |N/A |N/A |N/A |N/A |N/A |1.4        | ×  | ×  |1.2 |
 |WWWDG         | √  | ×  | √  | ×  | ×  | √         | ×  | ×  | √  |
 |**chxxxhw.h** | √  | ×  | √  | √  | √  | √         | ×  | ×  | √  |
-|**minichlink**| √  | ×  | √  | √  | √  | √  ( √   )| ×  | ×  | √  |
+|**minichlink**| √  | ×  | +  | √  | √  | √  ( √   )| √  | ×  | √  |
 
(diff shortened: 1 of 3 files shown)


monte-monte:
- Need to find linux terminal that doesn't support ANSI escape chars to test.
- Maybe this should be moved to a function in ``terminalhelp.h``?

cnlohr:
You know, you can call `system("")` to put cmd or powershell into ansi mode, right?

monte-monte:
No, I don't know that. What is that?

cnlohr:
It's how you get ANSI control codes in cmd and powershell.  I don't know why it's not better optimized.  Just call that at program start.  There are no downsides that I am aware of.

monte-monte:
weird, I didn't see any mention of it while I was looking for a way to enable ANSI in cmd on windows. Also that won't work on <win10? By using ``SetConsoleMode`` you can see if it works or no, and act accordingly.

cnlohr:
That is true, it is Win10+. But it will work in all situations/combinations on Win10+, so no need to check

monte-monte:
So, you're ok with making it incompatible with anything lower than win10? What's wrong with the proposed solution, that should cover all edge cases?

cnlohr:
`system("")` and I don't know.  It was told to me by someone who works at Microsoft. 

If you want you could also set with the set console properties!  I definitely don't want to deliberately leave out Win7 and lower.

PR #555, opened 2025-03-10
Fix PLL_MULTIPLICATION definition for ch32v303

miggazElquez:
The problem I noticed in #498 was only partially solved by 5ec73e145b374778156b2ef0a5c99666ae95417c: there is still a condition checking only for ch32v20x before setting `PLL_MULTIPLICATION`.

As this file is only included when using a ch32v30x I just removed the check.

### DIFF
--- a/ch32fun/ch32v30xhw.h
+++ b/ch32fun/ch32v30xhw.h
@@ -10436,7 +10436,7 @@ typedef struct
 				#define PLL_MULTIPLICATION RCC_PLLMULL15
 			#elif FUNCONF_PLL_MULTIPLIER == 16
 				#define PLL_MULTIPLICATION RCC_PLLMULL16
-			#elif defined(CH32V20x) && FUNCONF_PLL_MULTIPLIER == 18
+			#elif FUNCONF_PLL_MULTIPLIER == 18
 				#define PLL_MULTIPLICATION RCC_PLLMULL18
 			#else
 				#error "Invalid PLL multiplier"


cnlohr:
Err realized I can ping you here, @AlexanderMandera can you read over this and make sure it matches your intent?

AlexanderMandera:
@cnlohr
The change is fine to me, might be something I have overlooked initially.
It might be good to strip out the other if macros as well that check for unrelated chip families.

For example: https://github.com/cnlohr/ch32fun/blob/1a0df72b085da09b189f2b92408ea6b3761996d7/ch32fun/ch32v30xhw.h#L10408

could have the check for CH32V10x and CH32V20x removed.
Maybe we should do this for the other header files too.

cnlohr:
I would love to have another PR with that fix! If anyone else is up for testing it.  I just am not set up at the moment to test it.   Will merge this now.

miggazElquez:
Yeah, there is still quite a lot of checks for `ch32v20x` and `ch32v10x` in this file (and the same problem reversed in other files).
I can try deleting these for the 30x, but I'm testing on Zephyr, which doesn't support a lot of devices currently so I can only test if it build and can run Zephyr basic samples (like "Hello world").
I can also only test the ch32v303, but it seems to be the one the most affected by this problem (`#if defined(CH32V20x) || defined(CH32V30x_D8) || defined(CH32V10x)`. There is also some `#if defined(CH32V20x_D8) || defined(CH32V20x_D8W)` that I suppose could be removed ?

cnlohr:
IIRC there are different CH32V20x's so they could be removed from the v303 headers, but cannot be removed from the v20x headers.

PR #562, opened 2025-03-13
CH59x SysTick IRQ example

biemster:
Example how to use SysTick on CH59x to generate periodic interrupts for timing.


cnlohr:
Wonderful.  Everything looks good.

PR #563, opened 2025-03-14
CH59x UART1 printf and uartdemo example

biemster:
This example does `printf` on UART1 (rx,tx:PA8,PA9).

I #defined SetupUART like was done for ch32..., but I suspect that `_write` and `putchar` should be overloaded because of `WEAK`. I can change that easily if necessary.

Tested and working awesomely:
```bash
✔ ~/temp/CH570/ch32fun [ch59x-uartdemo L|✔] 
19:03 $ screen /dev/ttyUSB0 115200
ch32fun is awesome!
ch32fun is awesome!
ch32fun is aPA9 tx
PA9 tx
PA9 tx
PA9 tx
ch32fun is awesome!
ch32fun is awesome!
ch32fun
                                                                                                                                                                                                                  
[screen is terminating]
✔ ~/temp/CH570/ch32fun [ch59x-uartdemo L|✔] 
19:04 $ 
```

Also there are a bit more #defines in ch59xhw.h than necessary, but UART1 was mentioned in the interrupt from sleep stuff that I will be working on next so I included the whole sleep mode related #defines.


cnlohr:
Looks good.

PR #564, opened 2025-03-14
Fixes for 00x chips to build

dwillmore:
This is the start of my integration of the complete build system for the CH32x00x chips.  It includes changes to the global .mk and .h files as well as a small enhancement to the x00x specific header file.


dwillmore:
Holy whitespace changes, Batman!  Taking out one level of indent for a good chunk of the makefile.  Nothing changed past the 003/00x section except for a no longer needed endif.

dwillmore:
Can we merge this?

PR #566, opened 2025-03-15
[PROPOSAL] Set machine mode at startup

hexaedron:
The current way `mstatus` being initialized prevents us from manipulation of `mtvec` register. If you want to make your own custom bootloader, you need to alter `mtvec` right after the bootloader loads your firmware.

For example, i have my booloader starting from `0x08000000`:

```c
#define MAIN_CODE_ADDR (0x400);

int main()
{ 
    SystemInit();

    doSomeBootloaderStuffHere();

    void (*jump_func)(void) ;

    jump_func = (void (*)(void))MAIN_CODE_ADDR;

    jump_func(); 
}
```

Then my main firmware, which lives in `0x08000400`, takes control. The main firmware has its own startup code, so it tries to alter the mtvec register. If MCU is set to user mode in bootloader startup, i get a hardfault. If i initialize `mstatus` in the proposed way, everything works fine.

So my proposal is not to use user mode at all, but use machine node instead. Maybe, we should make this behaviour configurable in `funconfig.h`?

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -1128,10 +1128,10 @@ void handle_reset( void )
 	csrw 0xbc0, t0\n"
 
 #if defined(CH32V30x) && !defined( DISABLED_FLOAT )
-"	li t0, 0x6088\n\
+"	li t0, 0x7888\n\
 	csrs mstatus, t0\n"
 #else
-"	li t0, 0x88\n\
+"	li t0, 0x1888\n\
 	csrs mstatus, t0\n"
 #endif
 
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -1018,12 +1018,12 @@ void handle_reset( void )
 	// Setup the interrupt vector, processor status and INTSYSCR.
 
 #if FUNCONF_ENABLE_HPE	// Enabled nested and hardware (HPE) stack, since it's really good on the x035.
-"	li t0, 0x88\n\
+"	li t0, 0x1888\n\
 	csrs mstatus, t0\n"
 "	li t0, 0x0b\n\
 	csrw 0x804, t0\n"
 #else
-"	li a0, 0x80\n\
+"	li a0, 0x1880\n\
 	csrw mstatus, a0\n"
 #endif
 "	li a3, 0x3\n\


cnlohr:
I am a little confused... Why can't mstatus be updated by main later?  I'm definitely not against this, just wondering if it can be done easily in user code later.

TommyMurphyTM1234:
Some similar confusion over machine versus user mode arose here: 

- https://github.com/cnlohr/ch32fun/issues/479

I would have expected the ch32fun stuff to boot into and stay in machine mode since it's aimed at bare metal users. Does it actually switch to user mode early on which means that machine mode CSRs will no longer be directly available without switching back to the higher privilege machine mode first?

cnlohr:
So, MPP=3 = M-mode?  I've always struggled with that.

If so, I think we should make this change.

TommyMurphyTM1234:
Isn't it the case that the chip boots into machine mode? So if you do nothing at all with regard to privilege mode in `mstatus` then it simply stays in machine mode?

cnlohr:
> Isn't it the case that the chip boots into machine mode? So if you do nothing at all with regard to privilege mode in `mstatus` then it simply stays in machine mode? If the ch32fun startup code absolutely needs to change bits in `mstatus` for other purposes then maybe it should be masking them in/out as appropriate or doing a read/modify/write(back) and leaving other bits unchanged?

We do not `|=` / `&=` by default (when it makes sense not to do that) because it takes up more code space to do the `|=` / `&=` instead of direct assignment.  So, we should pick sane values and set them

hexaedron:
Yes @cnlohr  `MPP=11b` is machine mode. I just don't see any reason to set a user mode at startup but that's up to you.

hexaedron:
@cnlohr @TommyMurphyTM1234 Gyus please note my PR is about 10x, 20x and 30x. I have no idea about machine mode on 003, 03x etc.

cnlohr:
Would you mind adding machine mode to the 003 startup code and adding it to this PR?  I can test it later.

hexaedron:
I'll do that today.

cnlohr:
Seems not to cause any ill effects so I merged.

cnlohr:
(Other than 4 additional bytes required for startup code)

PR #569, opened 2025-03-17
CH59x Low Power and RTC demo

biemster:
The BLE chips have low power modes to be able to run on a small battery for very long. This PR adds a RTC demo, since the RTC is used to wake up the sleeping mcu, and a lowpower example that demonstrates how to use the low power sleep.

This PR requires a little help from linker script and assembly experts, since the interrupt vector needs to be in RAM in order to survive sleep. I managed to create a code section called "highcode" that can keep functions in ram that need to be available after wakeup, but the interrupt vector is more work.

### DIFF
--- a/examples_ch59x/lowpower/lowpower.c
+++ b/examples_ch59x/lowpower/lowpower.c
@@ -1,0 +1,45 @@
+#include "ch32fun.h"
+#include <stdio.h>
+
+#define LED PA8
+#define LED_ON FUN_LOW
+#define LED_OFF FUN_HIGH
+#define SLEEPTIME_MS 300
+
+__HIGH_CODE
+__INTERRUPT
+void RTC_IRQHandler(void)
+{
+	// clear timer and trigger flags
+	R8_RTC_FLAG_CTRL = (RB_RTC_TMR_CLR | RB_RTC_TRIG_CLR);
+}
+
+__HIGH_CODE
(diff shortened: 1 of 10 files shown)


examples_ch59x/lowpower/lowpower.c
```diff
@@ -0,0 +1,45 @@
...
+		Delay_Ms( 33 );
+	}
+}
+
+__HIGH_CODE
+int main()
+{
+	SystemInit();
+	RTCInit();
+    NVIC_EnableIRQ(RTC_IRQn);
```
cnlohr, line 32: Space/Tabs.

examples_ch59x/lowpower/lowpower.c
```diff
@@ -0,0 +1,45 @@
...
+    DCDCEnable();
```
cnlohr, line 33: Please add comments for each one of these commands.

cnlohr:
Your examples need more in-line comments for the calls they are making.  Please explain what you are doing more.

biemster:
I actually made https://github.com/biemster/ch592_minimal to test this better, but then I also immediately figured out that the code there works on fun as well.. All is tested and working now.

cnlohr:
I am glad to see these incremental quality improvements.  Thank you for your continued work on the ch592 port for ch32fun.

Issue #565, opened 2025-03-15
CH59x printf does weird stuff on strings longer than 7 chars


biemster:
Apparently I did not test this well enough, this needs fixing. Just putcharring with a delay works, so probably printf does it too quick?

TommyMurphyTM1234:
Perhaps it needs some sort of handshake/checking like the non-CH59x case? 

- https://github.com/cnlohr/ch32fun/blob/f1b8b67120886d47edbb2ad26ba6696a4985ae3d/ch32fun/ch32fun.c#L1319-L1329

biemster:
> Perhaps it needs some sort of "ready for transmit" handshake/checking like the non-CH59x case?

that is very possible, I just copied this from the EVT. Very possible that I missed something like this, thanks!

cnlohr:
![Image](https://github.com/user-attachments/assets/e91ad46d-199f-418a-af28-2efcbda84283)

cnlohr:
![Image](https://github.com/user-attachments/assets/1094d18f-c6eb-49ca-832b-490f5771d880)

biemster:
this was fixed in #570 (i thought that gh would close this automatically after merge)

PR #570, opened 2025-03-17
CH59x fix UART long printf issue #565

biemster:
For uart tx we should wait for the buffer to be empty so we don't risk overfilling it and dropping characters.


cnlohr:
Thanks

PR #571, opened 2025-03-19
Fixed ch32v003 random number example with native Linux and screen.

TheFlemoid:
Just started working with this library yesterday, but wanted to throw up a PR for this in case what I have found is a problem and not the norm.

I'm running Ubuntu 24 native, trying to program the CH32V003F4P6, using screen as my tty uART utility.  What I found was that the header definitions in the "random number" example didn't work as described, and I had to modify them to be more in line with the uartdemo header definitions.  Additionally, simply returning a newline character resulted in suboptimal output from screen for me, which was rectified by switching to a newline followed by a carriage return character.

I am fairly positive that these are all things that people have dealt with, cause this is a fairly popular library and a fairly popular MCU, but I at least wanted to get my opinions out while they were fresh on my mind in case they were valuable.  Feel free to tell me I'm dumb and close this as a no fix, I won't be offended.

### DIFF
--- a/examples/random_numbers/funconfig.h
+++ b/examples/random_numbers/funconfig.h
@@ -4,5 +4,9 @@
 #define RANDOM_STRENGTH    2
 #define CH32V003           1
 
+#define FUNCONF_USE_DEBUGPRINTF 0
+#define FUNCONF_USE_UARTPRINTF  1
+#define FUNCONF_UART_PRINTF_BAUD 115200
+
 #endif
 
--- a/examples/random_numbers/random_numbers.c
+++ b/examples/random_numbers/random_numbers.c
@@ -48,7 +48,7 @@ int main()
 	// * 3045866432
 	for(uint8_t x = 0; x < 4; x++)
 	{
-		printf("Random Number %u: %lu\n", x, rand());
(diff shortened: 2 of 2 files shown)


cnlohr:
We always recommend debug-over-swd over debug-over-uart for all users in all situations by default.  There are obviously situations where debug-over-uart is better, but, it is not the default.  The default run line should be `make clean all monitor` which will allow debug printf without any additional wires or anything.

As far as the spurious \r.  Yes, we should remove that.

TheFlemoid:
I figured there was something I was missing with the debug statements 😑 sorry about that!

I really appreciate your response (and your amazing toolchain here).  If it's okay with you, I'm going to go ahead and close this PR since 80% of it isn't relevant.  I do think going through and adding return carriages to examples that use the uart is probably a good idea (and I'm more then happy to go ahead and do that if you think it's worthwhile, since I found it a helpful catch), but that's not really what the PR was started for, so I wasn't sure if you'd want it open. 

Thank you for making this fantastic toolchain btw!

cnlohr:
Yes, feel free to close.

And thank you for the kind words. I still hope to keep this under steady improvement.

PR #573, opened 2025-03-22
minichlink: add ch643

recalci:
Every example from ch32x035 can run smoothly on ch643.


cnlohr:
Thank you for doing this work.

PR #574, opened 2025-03-22
CH59x: initial support in minichlink through "unknown-chip-type"

biemster:
This PR is to get basic support for ch59x in minichlink. Flashing works and terminal as well so it's quite usable like this already. Further research is needed to get this chip a first class citizen in minichlink but I think it's good to get this out so people can already use the basic functionality.

### DIFF
--- a/minichlink/pgm-wch-linke.c
+++ b/minichlink/pgm-wch-linke.c
@@ -385,7 +385,7 @@ static int LESetupInterface( void * d )
 			break;
 	}
 
-	// TODO: What in the world is this?  It doesn't appear to be needed.
+	// Set CLK Speed, 0x0C (0x01: 6000kHz, 0x02: 4000kHz, 0x03: 400kHz)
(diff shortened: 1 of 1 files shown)


biemster:
So the terminal works like this, but flashing not. For the time being flashing over swdio needs to be done using `wlink`.

cnlohr:
Any chance you could run this to ground?  Chances are it's some strange combination of the flash and block size controls.

biemster:
biemster/ch592_minimal can write to flash now, I will make an example with this in ch32fun shortly. After that we can try to figure out how to do that with direct register writes in the debugger.

cnlohr:
@biemster Do you want to do anything with this draft?

biemster:
> @biemster Do you want to do anything with this draft?

effort is picking up in this area, mostly with monte joining. I think this can be closed for now though since it will not be used like this.

PR #578, opened 2025-03-30
Add minichlink support for L103

recalci:
This commit adds minichlink support for CH32L103.

I noticed that minichlink isn’t compatible with CH32V103 when using either WCH-LinkE (CH32V305) or WCH-Link (CH549), so I marked it as "Work in Progress." Did I miss anything?


cnlohr:
I will commit this, but can you also make the changes here: https://github.com/cnlohr/rv003usb/blob/master/rvswdio_programmer/bitbang_rvswdio.h

PR #579, opened 2025-03-31
CH59x: add flashtest demo

biemster:
This is not working in fun, although the exact same code works in https://github.com/biemster/ch592_minimal.

I need an extra pair of eyes on this because I don't see anymore what the issue can be. I included libISP5xx temporarily to show that it should work, if you want to work on this that's the blob that should be RE'd.

One obviously wrong bit is the `while(!b)` in `flash_cmd_rom_write`, in the blob it is `while(b)`. But only like this it works in `_minimal` so I put it here to be fixed as well.

In `_minimal` this code has to be in `.highcode` to work, and in the blob it's there as well. I don't know why, it might be beneficial to figure out the reason and decide if that's the way we want to go too.

The main reason to add this demo is to aid #574 , to implement the direct method to flash new firmware in `minichlink`.

To finalize this demo after it's fixed the following still needs to be done:
- [ ] Move the extra functions to `ch59xhw.h`
- [ ] remove the ISP blob


cnlohr:
Should this be closed at this point in lieu of @monte-monte 's coming fixes?

biemster:
> Should this be closed at this point in lieu of @monte-monte 's coming fixes?

I think this is still very useful to have in examples_ch5xx, (and in the `hw.h` files), but it is still not working like this (although monte mentioned that just removing the .highcode attribute fixes it, I'll have a look at this again when isler rx works and the ch5xx SFRs are fully copied in)

biemster:
> If you don't mind I want to write this into a lib and put in in extralibs. Then will make it into a working example. 

That's an excellent idea, this one can be closed then.

PR #580, opened 2025-04-01
Added Register descriptions beside each peripherals struct

DevinJM3:
Copy and pasted the register description of each from each peripheral in the datasheet. This can and will make it easier to know what each register is/does without needing to check the datasheet each and every time

### DIFF
--- a/ch32fun/ch32v003hw.h
+++ b/ch32fun/ch32v003hw.h
@@ -93,11 +93,11 @@ typedef enum IRQn
 /* memory mapped structure for SysTick */
 typedef struct
 {
-    __IO uint32_t CTLR;
-    __IO uint32_t SR;
-    __IO uint32_t CNT;
+    __IO uint32_t CTLR;     /* System Count Control Register          */
+    __IO uint32_t SR;       /* System Count Status Register           */
+    __IO uint32_t CNT;      /* System Counter Register                */
     uint32_t RESERVED0;
-    __IO uint32_t CMP;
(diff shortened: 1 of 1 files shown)


biemster:
I did do the same when I added `ch59xhw.h`, so in my head this is clearly useful somehow. However I'd prefer to see this done for all `hw.h` files then, or none, to keep it uniform.

steveschnepp:
Also, a link to the datasheet at the beginning for the file might be a good addition ;)

DevinJM3:
> Nice!
> 
> I always like self-documented header files!
> 
> Small remark : can IDE pickup the description if it is defined after the field ? It would be really bad if IDE show the wrong description.

I assume it works the same with other IDE's or text editors, but it works perfectly in VSCode
image

Ill add a link to the datasheet and when I get a chance I will attempt to add the it to the other hw files

biemster:
> I assume it works the same with other IDE's or text editors, but it works perfectly in VSCode

I'm not an IDE user, but this indeed does look awesome

TommyMurphyTM1234:
> I assume it works the same with other IDE's or text editors, but it works perfectly in VSCode image

Is the `<unnamed>` because there's no struct name specified here? 
```
/* TIM */
typedef struct
{
    ...
} 
```
Would it make sense to name the structs to help with this?

cnlohr:
I am good with this as-is.  Any improvements would be great, though!

PR #581, opened 2025-04-05
Add support for 0.42 inch OLED

monte-monte:
Took settings from u8g2 https://github.com/olikraus/U8g2_Arduino/blob/master/src/clib/u8x8_d_ssd1306_72x40.c


cnlohr:
Thanks

PR #582, opened 2025-04-06
CH58x bringup and blinky, debugprintf demo

biemster:
The same porting procedure as for CH59x was used, which was also applied and working for CH573.
Fully validated on CH582F.

I did not aim for this PR number, I guess it was just meant to be.


cnlohr:
This is really good.  Only one change.  Is there any way to combine the ch5xx demos into one example folder?

PR #584, opened 2025-04-13
Improve buffered terminal in minichlink

monte-monte:
Presented changes are still guarded by ifdefs and will have effect only when compiled with ``TERMINAL_INPUT_BUFFER 1``.
- Fixed broken prints when incoming string had new lines anywhere inside of it but not at the end
- Fixed ``Send:`` label not showing until first terminal activity
- Added detection of ANSI compatibility of used terminal
- Fixed acking in ``b003boot`` terminal (not connected to the rest of the changes, but it's just a one line, didn't want to create a separate PR for it)

### DIFF
--- a/minichlink/minichlink.c
+++ b/minichlink/minichlink.c
@@ -519,18 +519,53 @@ keep_going:
 #if TERMINAL_INPUT_BUFFER
 							if ( nice_terminal )
 							{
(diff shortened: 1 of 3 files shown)


cnlohr:
Looks reasonable

PR #585, opened 2025-04-13
Improvements to b003boot

monte-monte:
- Added reboot into bootloader function: if no ``b003`` device is present, but we find a ``d003`` device, we send a magic reboot command and wait if it reboots. If during 5 seconds timeout it detects ``b003`` device it proceeds, else it tells that we couldn't establish connection to the bootloader
- Add new boot blob: this blob is backward compatible with old versions of rv003usb bootloader, but also fixes enumeration issue of boards without DPU pin, when used with newer version of bootloader. Blob array is left uncompacted to enable easy edits if we would need any. Later can be rewritten in more compact manner.


minichlink/pgm-b003fun.c
```diff
@@ -780,14 +840,50 @@ void * TryInit_B003Fun(uint32_t id)
...
-/* Run app blob
+/* Run app blob (old)
 				FLASH->BOOT_MODEKEYR = FLASH_KEY1;
 				FLASH->BOOT_MODEKEYR = FLASH_KEY2;
 				FLASH->STATR = 0; // 1<<14 is zero, so, boot user code.
 				FLASH->CTLR = CR_LOCK_Set;
 				PFIC->SCTLR = 1<<31;
 */
 
+/* Run app blob (new)
```
cnlohr, line 851: Can you add this inline with the text above the way you've written it?

minichlink/pgm-b003fun.c
```diff
@@ -96,14 +97,47 @@ static const unsigned char halt_wait_blob[] = {
...
+// 	0x23, 0xa6, 0x07, 0x00, 0x13, 0x07, 0x00, 0x08, 0x98, 0xcb, 0xb7, 0xf7,
+// 	0x00, 0xe0, 0x37, 0x07, 0x00, 0x80, 0x23, 0xa8, 0xe7, 0xd0, 0x82, 0x80,
+// };
+
 static const unsigned char run_app_blob[] = {
-	0x37, 0x07, 0x67, 0x45, 0xb7, 0x27, 0x02, 0x40, 0x13, 0x07, 0x37, 0x12,
-	0x98, 0xd7, 0x37, 0x97, 0xef, 0xcd, 0x13, 0x07, 0xb7, 0x9a, 0x98, 0xd7,
-	0x23, 0xa6, 0x07, 0x00, 0x13, 0x07, 0x00, 0x08, 0x98, 0xcb, 0xb7, 0xf7,
-	0x00, 0xe0, 0x37, 0x07, 0x00, 0x80, 0x23, 0xa8, 0xe7, 0xd0, 0x82, 0x80,
+	0xb7,0xf5,0xff,0x1f,
```
cnlohr, line 109: Please write in what these instructions are if you can.  It would help a lot of people who come after.

monte-monte: Do you mean describe by words, or in assembly? Because I've added assembly as a comment at the end of the file, where the translations of other blobs live.

cnlohr:
Github says you updated this ticket, but it doesn't say what it was.

cnlohr:
This is wonderful!! Thanks so much! I will merge when CI is done.

Issue #583, opened 2025-04-07
strchr - Bug or Feature ?


jjflash65:
In first - like ever - please excuse my bad english, it's neither good nor nice cause my native language isn't english.

I'm relatively new to CH32V003 and I'm evaluating possibilities of this chip. So I wanted to port some of my sources from STM32 and STM8 to CH32.

To my great surprise, the relatively simple source codes did not work. Namely those codes who are using strchr which are declared in <string.h>

I have tested the following program on every platform I use:

<pre>
  #include < string.h >

  int main(void)
  {
    char string[100];
    char *ptr, c = 'r';

    strcpy(string, "This is a string");
    ptr = strchr(string, c);
    if (ptr)
      printf("The character %c is at position: %d\n\r", c, ptr-string);
    else
      printf("The character was not found\n\r");

    while(1);
  }
</pre>

Every time the output was:

`The character was not found`

With CH32FUN the output is:

`The character z is at position: 16`

An original string.h discription says:

<b>char *strchr(const char *string, int c);</b>

The strchr() function finds the first occurrence of a character in a string. The character c can be the null character (\0); the ending null character of string is included in the search.

The strchr() function operates on null-ended strings. The string arguments to the function should contain a null character (\0) that marks the end of the string.

<b>Return Value</b>
The strchr() function returns a pointer to the first occurrence of c that is converted to a character in string. The function returns NULL if the specified character is not found.

In ch32fun.c
https://github.com/cnlohr/ch32fun/blob/master/ch32fun/ch32fun.c#L384

I found the source of strchr. This code finds the first occurrence of c, but if the searched char isn't present, the function returns pointed to the zero-byte at the end of the string (and not NULL).

For me I had replaced the function with this:

<pre>
/* ----------------------------------------------------------
                           strch
                           
     The strchr() function returns a pointer to the first 
     occurrence of ch that is converted to a character in
     string. The function returns NULL if the specified 
     character is not found.
   ---------------------------------------------------------- */

char *strch(const char *str, char ch)
{
  while(*str)
  {
    if (*str == ch) { return (char *)str; }
    str++;
  }
  if ((*str == 0) && (ch== '\0')) return (char *)str;
  return NULL;
}</pre>

Maybe you want to replace it too ?

TommyMurphyTM1234:
> ```
> char string[15];
> ...
> strcpy(string, "This is a string");
> ```

[...]

PR #587, opened 2025-04-15
Fix implementation of strchr, ref #583

cnlohr:
Still need to test this and need sign-off by @jjflash65


cnlohr:
Err the comment doesn't look like it accepted.  @jjflash65 can you test this and sign off?

jjflash65:
I have tested the code and it works fine for me. I have closed the issue !
Thanks for fixing !

cnlohr:
Thank you.

PR #590, opened 2025-04-24
add rtc example

Outfisken:
This example demonstrates how to configure the RTC and set an alarm interrupt.

### DIFF
--- a/examples_v30x/rtc/Makefile
+++ b/examples_v30x/rtc/Makefile
@@ -1,0 +1,10 @@
+all : flash
+
+TARGET:=rtc
+TARGET_MCU:=CH32V303
+TARGET_MCU_PACKAGE:=CH32V303CBT6
+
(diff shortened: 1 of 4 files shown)


cnlohr:
Adding some information to the README.md surrounding how you tested and some results would be hugely helpful to people if you are willing to add it.  If not, I am willing to accept as-is.

Information like:
1. Does this apply to all v30x family parts?
2. Are there any caveates or connection concerns? 
3. What should be done with things like VBAT?  Connect to VCC, or with diode and capacitor?
4. What did you observe for power usage when just using RTC?

Outfisken:
Tried to measure the power consumption, but it appears to be less than 1uA. Unfortunately, I dont have any hardware to perform any measurements in that range.

cnlohr:
Thanks!  That sounds great.  And the added documentation is very helpful.

PR #598, opened 2025-05-14
Update v003, x03x, v103, v20x

dwillmore:
I changed and edited the comment styles used in the examples to make them consistent.  
I also picked new GPIO for the different chips with an eye towards pins that all packages have in common and are outputs as their primary function.

Moved to symbolic pin names.
Moved to fun* function interface instead of raw register manipulation.


dwillmore:
Tested on 003, x033, v203.  

I do not have a v103 to test on.  Requesting volunteers for that as it was heavily edited because it was very old.

Also, included bin files for those who need a quick start before they have the development environment fully up and running.

dwillmore:
30x and the CH5[789]x are next.  

I'll need a 30x tester as well.

PR #599, opened 2025-05-15
CH5xx examples consolidation

biemster:
The examples for all the ch5xx are very similar, and with a bit of effort every single example should work on the whole range of chips. Since the ch59x had the most examples, this change will remove the ch57x and ch58x examples and change the ch59x examples into ch5xx.

They will be tested on ch570, 573, 582 and 592, and default to 570 since that's the new kid on the block.


biemster:
This is done! woohoo!

examples_ch5xx/blink/blink.c
```diff
@@ -0,0 +1,35 @@
...
+#define PIN_1 PA4
+#define PIN_K PA5
+#define PIN_BOB PA8
+#define PIN_KEVIN PA9
+
+int main()
+{
+	SystemInit();
+
+	funGpioInitAll(); // Enable GPIOs
```
cnlohr, line 14: Only thing I would add here is that depending on the chip, this may be a no-op.

examples_ch5xx/uartdemo/uartdemo.c
```diff
@@ -5,10 +5,12 @@ int main()
...
 	SystemInit();
 
+	funGpioInitAll(); // Enable GPIOs
+
 	funPinMode( PB22, GPIO_CFGLR_IN_PU );
 
-	SetupUART(FUNCONF_UART_PRINTF_BAUD); // UART1 on rx,tx:PA8,PA9 at 115200 baud
-	printf("TX in on PA9, RX is on PA8 but not implemented in this demo.\r\n");
+	SetupUART(FUNCONF_UART_PRINTF_BAUD); // UART1 on rx,tx:PA8,PA9 at 115200 baud (PA2,PA3 on ch570/2)
+	printf("TX in on PA3 or PA9, RX is on PA2 or PA8 but not implemented in this demo.\r\n");
```
cnlohr, line 13: Please write which IOs it's on on which platforms.

examples_ch5xx/blink/blink.c
biemster: all of ch5xx is no-op, I'll add that

examples_ch5xx/uartdemo/uartdemo.c
biemster: the comment actually has this info, it's just the printf that is generic. I did not want to put ifdefs around it. This was mostly to test printing of long strings, I'll come up with some generic that is not ambiguous like this.

cnlohr: Not worth ifdeffing.  Thanks for the extra comment.

examples_ch5xx/blink/blink.c
cnlohr: Thanks!

biemster:
I was almost done with the comments you added!

PR #602, opened 2025-05-21
CH5xx SFR completion

biemster:
While adding support for the ch5xx range one by one, I only added the parts of the SFR that were needed for the examples that were implemented. This change package will add the rest of the SFR pointer #defines so all the peripherals are accessible, even if they don't have an example in the `examples_ch5xx` folder.


biemster:
this is ready to merge @cnlohr , and will conclude the ch5xx consolidation! :tada:

cnlohr:
Thanks for the work.

PR #605, opened 2025-05-24
Add 64bit SysTick example

BogdanTheGeek:
Modified the original example to work on split 64bit counters.
You can see which other MCUs this would work on here:
https://github.com/cnlohr/ch32fun/blob/c234dad6a4715651952c1f4a0fff2b8d379cddde/ch32fun/ch32fun.c#L1484

I guess alternatively we could change the typedef for these to have a single 64bit register. They are little endian, so it should work.


cnlohr:
It's very frustrating that there is no 32-bit mode :( but I guess we have to do what we have to do.

PR #608, opened 2025-05-28
Cleanup x035 declarations and USBPD

BogdanTheGeek:
Some things were misspelled or had an inconsistent style.
I also removed all of the implementation specific USB PD stuff as they have nothing to do with the hardware.
Lastly, I redefined UPD_TMR_* to be calculated automatically instead of having to use #if .. #elseif ...
I've implemented a single header USB PD library that works with these changes here: https://github.com/BogdanTheGeek/ch32fun-usbpd


cnlohr:
Looks great.  Thanks @BogdanTheGeek

PR #609, opened 2025-05-28
[minichlink] open serial port nonblocking

olell:
Opening the serial port in nonblocking mode and ignoring the carrier detect line allows it to open also serial devices which do not provide those lines (e.g. atmega32u4 USB CDC as ardulink). This PR adds the `O_NONBLOCK` flag, clears it right after opening the device and sets the `CLOCAL` attribute to ignore the CD-line.

I am not a hundred percent sure if this is only required on macOS, but I couldn't test if this is also required on linux or would even upset the linux build, maybe someone with an ardulink and linux machine could test it?

### DIFF
--- a/minichlink/serial_dev.c
+++ b/minichlink/serial_dev.c
@@ -65,8 +65,15 @@ int serial_dev_open(serial_dev_t *dev) {
 	}
 #else
 	struct termios attr;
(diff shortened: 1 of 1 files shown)


cnlohr:
This looks good, I don't use this mode but I am glad you are exploring and testing this.  Will merge when CI is done.

PR #616, opened 2025-06-01
RFC: Live Watch Server

BogdanTheGeek:
Based on the gdb stub implementation, I have added an openOCD style command port to allow reading and writing memory values while the processor is "running" (the implementation actually halts the processor during the read)
THIS IS NOT PRODUCTION QUALITY CODE!!!!!
Ive added `read.sh` as an example.


BogdanTheGeek:
closing so it doesn't keep running the workflows

BogdanTheGeek:
This POC is as "flexible" as you can imagine, maybe it would be better to just have a read/write regs "api", to decouple the hardware specific stuff (maybe we will be able to read memory without halting in the future)

PR #619, opened 2025-06-11
ch573/1: fix RTC_irq demo

biemster:
The interrupt vector of 571/3 consists of jump instructions, not plain addresses for the handlers. After adding this the RTC irq demo is also working for this chip.


cnlohr:
This looks amazing.  Thanks.

cnlohr:
I just YOLO'd it without a CI check cause you're 😎

PR #621, opened 2025-06-15
minichlink: add linkW

biemster:
This is just to have the linkW reported as a linkW and not as Unknown. No functional changes at all, in my limited testing it seems to work as well as a linkE.


cnlohr:
Thanks.

PR #622, opened 2025-06-21
Add spi_lsm6dsv32x example

AlexanderMandera:
Hi all,

I have created an example for the LSM6DSV32x IMU sensor, based on the bounty provided by spice-weasel:
https://discord.com/channels/665433554787893289/1384394865428926474/1384394865428926474

Can somebody also check if the license is correctly integrated in this example?
Parts of the ST library are used for this example, especially for the reading part, but also the lsm6dsv32x-pid library is licensed under BSD 3-clause. Maybe I need some advice here.

Is also using a submodule alright here?

I hope this example will benefit anyone using this new family of IMU sensors with CH32 series :)
Theoretically, with 7520 bytes of flash usage, it is even small enough to run on CH32V003 if you wanted to 👍

I am open to any suggestions.

Thanks
Alex

### DIFF
--- a/examples_v20x/spi_lsm6dsv32x_nolib/spi_lsm6dsv32x_nolib.c
+++ b/examples_v20x/spi_lsm6dsv32x_nolib/spi_lsm6dsv32x_nolib.c
@@ -1,0 +1,166 @@
+/*
+ * spi_lsm6dsv32x - Example for LSM6DSV32X sensor implementation using SPI interface
+ * with CH32V20x microcontroller. Created by Alexander Mandera 
+ *
+ * Portions of this file are derived from the STM32 LSM6DSV32X library,
+ * licensed by ST under BSD 3-Clause license.
+ *
+ * Copyright (c) 2021 STMicroelectronics.
+ * All rights reserved.
+ */
+
+#include "ch32fun.h"
+#include <stdio.h>
+#include <math.h>
+
+#include "lsm6dsv32x_spi.h"
+
+// Functions
+void initialize_lsm6dsv32x( void );
+void read_data( void );
+
+/*
+    Pinout for SPI:
+    - PA5: SCK (SPI clock)
+    - PA6: MISO (Master In Slave Out)
+    - PA7: MOSI (Master Out Slave In)
+    - PA4: CS (Chip Select, NSS pin for SPI1)
+*/
+
+uint8_t register_setup[] = {
+	LSM6DSV32X_CTRL3,			0b00000010, // CTRL3: Block Data Update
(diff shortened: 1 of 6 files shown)


cnlohr:
Many people don't want to use gitmodules.  Because the header files are licensed under non-MIT-licensing, you would need to just #define the pertinent registers yourself in a header file.

But also, I see you are using a lot more abstracted functions :confused: I wouldn't imagine you would want to rewrite this:

```c
    lsm6dsv32x_xl_data_rate_set( &dev_ctx, LSM6DSV32X_ODR_AT_7Hz5 );
    lsm6dsv32x_gy_data_rate_set( &dev_ctx, LSM6DSV32X_ODR_AT_15Hz );
```
to be
```c
    spi_write( FIFO_CTRL3, 0b00110010 ); // configure gyro = 0b0011 = 15Hz, accelerometer = 0b0010 = 7.5Hz
```

The `lsm6dsv32x-pid` is also a tad heavy for this sort of stuff, but not too bad, just a bummer it can't be "brought in" -- and even if it could, people would murder me for using a .gitmodule

side-note: the STM lib doing a callback with a function pointer adds a **a lot** of overhead.  I don't know why they would have written the abstraction this way, it would have been much thinner if they just called the function.

cnlohr:
Just to loop back with the discussion, it actually would be much better size-wise, to switch it to an initialization list.

I.e.
```c
uint8_t register_setup[] = {
  FIFO_CTRL3, 0b00110010, // configure gyro = 0b0011 = 15Hz, accelerometer = 0b0010 = 7.5Hz
  FIFO_CTRL2, ...
  };
  ```

File: examples_v20x/spi_lsm6dsv32x_nolib/spi_lsm6dsv32x_nolib.c [hunk 1]
```diff
@@ -0,0 +1,166 @@
...
+	GPIOA->CFGLR |= ( GPIO_Speed_50MHz | GPIO_CNF_OUT_PP_AF ) << ( 4 * 5 );
+
+	// SPI Master mode
+	SPI1->CTLR1 |= SPI_Mode_Master;
+
+	// Enable Full duplex mode
+	SPI1->CTLR1 |= SPI_Direction_2Lines_FullDuplex;
+
+	// MOSI on PA7, 10MHz Output, alt func, push-pull
+	GPIOA->CFGLR &= ~( 0xf << ( 4 * 7 ) ); // Clear PA7 config
```
cnlohr, line 74: You may want to use funPinMode here as well.

cnlohr:
Also need to fix CI!

PR #632, opened 2025-06-29
Add USB CDC-ACM example

monte-monte:
Also fixed hw file for CH32V103 and added SysTick example for it.


cnlohr:
For completeness (Writing here).
1. There is a bug when using loopback.
2. Needs to add the RX configuration to make it obvious to developers what pins are being used.
3. CH32V203 usage seems very buggy.  This needs to be commented out:
```c
#ifdef CH32V20x
			ret = -1; // ACK, without it RX is stuck on ch32v208, but is ok on ch32v307 (strange?)
#endif
```
4. Needs more text in main README.md explaining how it works + What pins do what on what processors.

PR #633, opened 2025-07-01
Add CDC ACM example, clean

monte-monte:
I will add more comments to the code and clean up dead parts when I will be adding CH5xx uart code. Also will try to write broader readme after that.


cnlohr:
image

PR #635, opened 2025-07-02
Fix funconf clarification

cnlohr:


cnlohr:
Reminder to self - if merging, squash commit.

PR #636, opened 2025-07-02
Add CAN Bus Example

BogdanTheGeek:
This example should also work on v20x.
Had to fix some of the flags as they didn't match the RM and would truncate the MSB of the extended ID.

### DIFF
--- a/examples_v30x/canbus_network/funconfig.h
+++ b/examples_v30x/canbus_network/funconfig.h
@@ -1,0 +1,7 @@
+#ifndef _FUNCONFIG_H
+#define _FUNCONFIG_H
+
+// #define CH32V30x           1
+
+#endif
+
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ minichlink/minichlink.so
 compile_commands.json
 .cache
 ch32fun/generated_*.ld
+**/.DS_Store
--- a/ch32fun/ch32v20xhw.h
+++ b/ch32fun/ch32v20xhw.h
@@ -2038,7 +2038,7 @@ typedef struct{
 #define CAN_TXMI0R_TXRQ                         ((uint32_t)0x00000001) /* Transmit Mailbox Request */
 #define CAN_TXMI0R_RTR                          ((uint32_t)0x00000002) /* Remote Transmission Request */
 #define CAN_TXMI0R_IDE                          ((uint32_t)0x00000004) /* Identifier Extension */
-#define CAN_TXMI0R_EXID                         ((uint32_t)0x001FFFF8) /* Extended Identifier */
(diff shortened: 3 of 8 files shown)


examples_v30x/canbus_network/funconfig.h
```diff
@@ -0,0 +1,7 @@
+#ifndef _FUNCONFIG_H
+#define _FUNCONFIG_H
+
+// #define CH32V30x           1
```
cnlohr: You can remove this.

cnlohr: You also should specify clock source/frequency/etc. here.  I don't even remember what default is.  but because it matters for you it should be specified.

biemster:
I get two "defined but not used" when compiling this, if we want to move the CI (or all builds?) to `-pedantic` as discussed in the minichlink ch5xx PR it will trip over this.

biemster:
Just a nit, but maybe you could add to the README why your simple PHY would be worse than a proper one? I'm quite tempted to use the one given here in some future project because of it's simplicity, but I probably hit some limitation earlier than I expect?

BogdanTheGeek:
> Just a nit, but maybe you could add to the README why your simple PHY would be worse than a proper one? I'm quite tempted to use the one given here in some future project because of it's simplicity, but I probably hit some limitation earlier than I expect?

Its just not compatible with any CAN PHY standard, you can use it just fine with others of the same kind for *fun* projects.

cnlohr:
> Its just not compatible with any CAN PHY standard, you can use it just fine with others of the same kind for fun projects.

Do you mean that you can't use it with regular / existing CAN networks?   If so, there needs to be big warnings.

biemster:
> > Its just not compatible with any CAN PHY standard, you can use it just fine with others of the same kind for fun projects.
> 
> Do you mean that you can't use it with regular / existing CAN networks? If so, there needs to be big warnings.

image

There is, the warning is fine, I was just wondering if there were other limitations over using a proper phy.

BogdanTheGeek:
finally got a can phy and a v208 and it works as expected, i will add some timeouts for tx and a way to make the baud rate macro a bit better, so it works with all prescalers.

cnlohr:
spectacular!! I would love a schematic so I can connect it to my car or something

cnlohr:
just let us know when the code is updated.

BogdanTheGeek:
> spectacular!! I would love a schematic so I can connect it to my car or something

image

Here's a simplified version of the 208 boards I made. 

WARNING: the TJA104xT ties VCC and VIO together internally (pin 5 is n.c.), which is fine for 20x because they have 5v tolerant io, but it may damage other ics. use the TJA104xBT if you need 3v3 io.

@cnlohr i think its ready to merge
I've also added a 20x example, so #638 can be closed.

cnlohr:
@BogdanTheGeek please include the schematic in the files for the PR.

PR #637, opened 2025-07-03
Fix USB definitions in ch32v30xhw.h

monte-monte:
There was a mistake in one reg definition.


cnlohr:
Thanks for catching it.

PR #641, opened 2025-07-17
iSLER

biemster:
This is the current state of iSLER as in https://github.com/biemster/ch32fun_isler.

All RF enabled MCU's have support at some level, most have full support. It is a bit messy due to all the `#defines`, but not more than the rest of `ch32fun`.

There is quite a bit more opportunity for experimentation, and very likely not all the register writes are necessary so it could be slimmed down, but it's good to have it out there so people can easily start improving it.

### DIFF
--- a/examples_ch5xx/README.md
+++ b/examples_ch5xx/README.md
@@ -8,9 +8,11 @@ The following table details which demo is working on which chip, since some func
 |-------------------|---------|---------|-------|-------|
 | blink             |    √    |    √    |   √   |   √   |
 | debugprintfdemo   |    √    |    √    |   √   |   √   |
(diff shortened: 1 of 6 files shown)


biemster:
Actually what most likely needs to be added before it's ready to be merged is taking out the AccessAddress from the header, because you usually don't want to send everything on the BED6 address, only BLE advertisements.

cnlohr:
@biemster do you plan to do that?  Other than that this looks ready!

biemster:
> @biemster do you plan to do that? Other than that this looks ready!

then let's get this out there! There will be more small PRs on this in the near future, the AccessAddress will be one of them then. :rocket:

PR #642, opened 2025-07-17
minichlink: builtin bootloader ISP programming

biemster:
This change package will enable `minichlink` to program chips using their builtin USB bootloader. This is mostly the ch5xx range, and enables only the `-w` write command (and `-b`) using the ISP interface. And the -E erase command.

I think this is really handy if one has to program a lot of new chips, since they don't need to separately enable the debug interface over USB and then flash the firmware, but can just immediately write the firmware.

This is also useful for people not owning a linkE (yet).

A future improvement will be the addition of `FUNCONF_USBDEBUGPRINTF`, which will get a feature to reboot the mcu into the internal bootloader so minichlink can be used like this without reconnecting the USB cable.

### DIFF
--- a/minichlink/Makefile
+++ b/minichlink/Makefile
@@ -1,7 +1,7 @@
 TOOLS:=minichlink minichlink.so
 
 CFLAGS:=-O0 -g3 -Wall -DCH32V003 -I. -DMINICHLINK
-C_S:=minichlink.c pgm-wch-linke.c pgm-esp32s2-ch32xx.c nhc-link042.c ardulink.c serial_dev.c pgm-b003fun.c minichgdb.c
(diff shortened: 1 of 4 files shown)


biemster:
This is tested and works on ch582 and ch59x, and maybe 570 as well. It does not work for ch573, but I undrafted anyway because I wanted the reviews to start. And it is ready as is, support will be extended in the future.

biemster:
I've added ch570/2 support, I won't touch it for a while now so it can be reviewed. This is basically all living in a new file, and the main minichlink code is changed very minimal so review should be quick and easy.

cnlohr:
@biemster this all seems relatively straightforward and fine.  I don't know why I wouldn't merge this if you've tested it?

biemster:
> @biemster this all seems relatively straightforward and fine. I don't know why I wouldn't merge this if you've tested it?

yes this is a very low impact merge, if you agree with the approach taken (treat the builtin bootloader as a programmer) please feel free to merge!

PR #644, opened 2025-07-25
minichlink: fix ch5xx ISP programming on non-zero offset

biemster:
The new ch5xx ISP programming mode had an issue where it did not respect the offset where the programming should start, but always programmed the binary at 0x00000000.
This change will fix that, and is tested on ch582.


cnlohr:
Is this good to commit @biemster

biemster:
> Is this good to commit @biemster

yes please and thank you!

PR #648, opened 2025-07-28
Add dma gpio mem2mem example

nolanhergert:
Clarify nuances of triggered dma gpio example and add a new example that runs at 8MS/sec using dma mem2mem mode


cnlohr:
Looks good.  Going to just check against CI with all the recent changes.

PR #654, opened 2025-08-01
USB CDC TTY and FUNCONF_USE_USBPRINTF

biemster:
This demo shows a simple printf like behavior, but over USB CDC. It'll create a `/dev/ttyACM0` on linux (not sure what it does on windows), where one can monitor the messages and send simple commands.

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -811,7 +811,7 @@ void PrintHex( uint32_t n )
 // If you don't override a specific handler, it will just spin forever.
 void DefaultIRQHandler( void )
 {
-#if FUNCONF_DEBUG_HARDFAULT && ( FUNCONF_USE_DEBUGPRINTF || FUNCONF_USE_UARTPRINTF )
+#if FUNCONF_DEBUG_HARDFAULT && ( FUNCONF_USE_DEBUGPRINTF || FUNCONF_USE_UARTPRINTF || FUNCONF_USE_USBPRINTF )
 	//This is kind of like a crash handler.
 	//printf( "DEAD MSTATUS:%08x MTVAL:%08x MCAUSE:%08x MEPC:%08x\n", (int)__get_MSTATUS(), (int)__get_MTVAL(), (int)__get_MCAUSE(), (int)__get_MEPC() );
 	PrintHex( __get_MEPC() ); // "addr2line -e debugprintfdemo.elf 0x000007e6" ---> debugprintfdemo.c:45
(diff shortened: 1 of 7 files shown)


biemster:
This PR is ready, but @monte-monte can you have a look before I undraft? I had to un-static `USBFS_SendEndpointNEW` in `usbfs.c`, not sure if you agree with that?

monte-monte:
Seems ok. I would add more configs for different chips, you can copy it from my example if you like.

biemster:
> Seems ok. I would add more configs for different chips, you can copy it from my example if you like.

in the makefile you mean? I don't have any more chips to test it with, so I was kind of hoping that others would test and add the config in a small PR.

monte-monte:
Makefile and funconfig.h
Since I've tested on all that, I think your example should work too. Just check the one from my latest PR (not yet merged) there are important additions for ch32v208
If someone find a bug they will report it :)
But if there are predefined config for multiple chips it will be easier for people to try it out.

biemster:
> Makefile and funconfig.h Since I've tested on all that, I think your example should work too. Just check the one from my latest PR (not yet merged) there are important additions for ch32v208 If someone find a bug they will report it :) But if there are predefined config for multiple chips it will be easier for people to try it out.

The `funconfig.h` stuff does not work on my machine, I think this gets included before the CH5xx macro in usb.c (or in a different gcc call?) so it does not pick up the chip specific things.

cnlohr:
> > Makefile and funconfig.h Since I've tested on all that, I think your example should work too. Just check the one from my latest PR (not yet merged) there are important additions for ch32v208 If someone find a bug they will report it :) But if there are predefined config for multiple chips it will be easier for people to try it out.
> 
> The `funconfig.h` stuff does not work on my machine, I think this gets included before the CH5xx macro in usb.c (or in a different gcc call?) so it does not pick up the chip specific things.

wat?

funconf.h needs to .. work.  What do you mean? your chip selection should be done in the Makefile, but if it's a feature configuration that... important.  can you poke at it?

biemster:
> > The `funconfig.h` stuff does not work on my machine, I think this gets included before the CH5xx macro in usb.c (or in a different gcc call?) so it does not pick up the chip specific things.
> 
> funconf.h needs to .. work. What do you mean? your chip selection should be done in the Makefile, but if it's a feature configuration that... important. can you poke at it?

The problem is a bit of trickery that monte tried in the usb_cdc_uart example, where a new ch5xx macro is defined in usbfs.c, and used in funconfig.h. This does not work since funconfig.h is included before the usb stuff, and thus ch5xx specific `FUNCONFIG_` macros are not set.
This is easily remedied by using the non-consolidated CH5{789}x macros in funconfig.h instead of the ch5xx one. I could add that now but maybe monte has other plans in the original usb_cdc_uart example so I'll wait for the discussion to conclude.

This PR can be merged just for CH5xx now, I'll make a small update PR where support is added for more chips then. (or wait with merging and I'll update this at that time, whichever you prefer.)

cnlohr:
Please try to make sure the update comes in.

PR #656, opened 2025-08-04
Add missing blink binaries

monte-monte:
Added missing blink binaries. For ch5xx there is a separate binary for ch585, because it has clock that is not compatible with other chips in the family.


cnlohr:
looks good

PR #658, opened 2025-08-08
Fix unbrick. Enhance prints for b003boot init.

monte-monte:
Now before unbricking we reset linkE's state. Without this step it fails with CH570/2.
Add prints for UUID and read protection status for b003boot to make it look inline with other programmers.


cnlohr:
Looks reasonable.

PR #659, opened 2025-08-08
BLE beacon low power demo

biemster:
This is an example of how to implement a simple low power BLE beacon with iSLER. While both low power and iSLER already worked on many chips, the issue was that after a low power state the RF would not properly wake up anymore.
This change package fixes that as well.


biemster:
This is complete, I'd like this out there so people with a PPK2 can do power tests on them. It is possible that the wakeup can be optimized, so it uses even less power.

cnlohr:
amazing :magic_wand:

Issue #660, opened 2025-08-10
blink does not work anymore with ardulink


zoobab:
Hi,

I am trying to flash a sop16 ch32v003 on a breakout board with an ardulink, but the blink example give this error (version 3abf32ba0bb1eb97ca4222cd05fa81cc3dd06369 of 8 aug):

```
../../ch32fun//../minichlink/minichlink -C ardulink -c /dev/ttyACM0 -w blink.bin flash -b
Opening serial port /dev/ttyACM0 at 115200 baud.
Ardulink: synced.
Interface Setup
Unknown chip type.  Report as bug with picture of chip.
marchid : ffffffff
HARTINFO: ffffffff
make: *** [../../ch32fun/ch32fun.mk:351 : cv_flash] Erreur 254
```

Now I tried my old personal fork of some months ago (see zoobab/ch32v003fun), and it works fine:

```
ch32v003fun/examples/blink$ ../../ch32v003fun/../minichlink/minichlink -c /dev/ttyACM0 -w blink.bin flash -b
Opening serial port /dev/ttyACM0 at 115200 baud.
Ardulink: synced.
Found Ardulink Programmer
Interface Setup
Image written.
```

zoobab:
I checked out 7e6e46e9d1e4b507f6d4e310b8e5e2497a02e53d , it still works:

```
../../ch32fun//../minichlink/minichlink -w blink.bin flash -C ardulink -c /dev/ttyACM0 -b
Opening serial port /dev/ttyACM0 at 115200 baud.
Ardulink: synced.
Interface Setup
Writing image
Chip Type: 071
Autodetected a SWDIO chip (Enum: 09 from 0713bb91)

Image written.
```

Let me try to pin point the culprit commit...

zoobab:
Still working with 533fedc4c30779b6a51cb0005ffe27f2e31ee2e2

cnlohr:
@monte-monte did you want to bring a discussion here so others can find if they search?

monte-monte:
As I said on discord, I need to built ardulink to debug with first. I never used it. It will take some time, but I hope to fix this this week.

cnlohr:
Wait, @monte-monte sorry, I just realized @BlueSyncLine is the one who contributed that, so they should be the first line of defense to fixing things up if possible.

monte-monte:
@cnlohr I'm confused, the report suggests it's after my changes.

monte-monte:
[...]

PR #665, opened 2025-08-12
Fix ardulink inititalization

monte-monte:
fixes https://github.com/cnlohr/ch32fun/issues/660


cnlohr:
Thanks for fixing this even though it was not your responsibility.

PR #667, opened 2025-08-13
Update ch32fun.mk to erase chip each time make flashes the device

AdonaiDiazEsparza:
Just add flag for full erase, maybe this could be also done with the make command but i would let that into discussion

### DIFF
--- a/ch32fun/ch32fun.mk
+++ b/ch32fun/ch32fun.mk
@@ -332,7 +332,7 @@ clangd_clean :
 	rm -f compile_commands.json
 	rm -rf .cache
 
-FLASH_COMMAND?=$(MINICHLINK)/minichlink -w $< $(WRITE_SECTION) -b
+FLASH_COMMAND?=$(MINICHLINK)/minichlink -E -w $< $(WRITE_SECTION) -b
 
 .PHONY : $(GENERATED_LD_FILE)
 $(GENERATED_LD_FILE) :


cnlohr:
Well, we don't want to change the default behavior for `flash` because people will put data partitions in other parts of their chip, or to store important data and we don't want to erase them first.  But, we could consider a new target, for instance `make reflash` or something like that.  

Would you mind join in our Discord?  you can reach me at `cnlohr` on discord and I can send you an invite.

AdonaiDiazEsparza:
Sure, I’ll accept the invitation to join your Discord.
This pull request was made because one of my CH32 chips started behaving strangely after reflashing using make. If you could add a make reflash option without changing the default make flash behavior, it would be useful for cases where someone needs to flash firmware to an erased memory.

cnlohr:
If you did want to alter this so it was a new target like `reflash` I'd say go for it.  I am just anxious to supersede existing functionality.

cnlohr:
You can also add a `reflash` rule.  Good idea moving it to cv_reflash though.

AdonaiDiazEsparza:
something like this, just to erase the chip and upload the current .bin file, right?
`reflash : 
	$(MINICHLINK)/minichlink -E -w $< $(WRITE_SECTION) -b`

cnlohr:
Yes, though, I don't believe many of the flashers know to make the flashing process go turbo when the chip is fully erased, they _could_

PR #668, opened 2025-08-13
Ch582 low power

biemster:
The ch582 does not have the low power functions implemented yet, and thus can't to battery powered BLE beacon yet.
This package implements the RTC irq demo, low power demo and ble beacon demo for this chip, making it one of the best supported chips in the ch5xx line together with ch570/2 and ch59x.


cnlohr:
I will let you decide when this is ready for review.

biemster:
This is done and working, and I tested the affected demos on ch570, ch582 and ch592.

PR #672, opened 2025-08-15
Fix udev rules on 99-minichlink.rules:

YuukiUmeta-UIAP:
- Change multiple GROUP configuration from udev codes to ACL because udev only allow a single GROUP definition for a single device
- Add fallback that dose nothing if setfacl commund doesn't exist, when ACL execute

No any effect with changing is verified on Ubuntu 24.04 LTS using `ls -lR` and `getfacl -R` following steps:
1. Before change
    - ls -lR /dev > dev_before.txt
    - getfacl -R /dev > dev_acl_before.txt
2. After change
    - ls -lR /dev > dev_after.txt
    - getfacl -R /dev > dev_acl_after.txt
3. Then, `diff` each other to verify there is any unintended changes

I think that this PR should be discussed carefully.

### DIFF
--- a/minichlink/99-minichlink.rules
+++ b/minichlink/99-minichlink.rules
@@ -1,28 +1,19 @@
-SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="8010", GROUP="plugdev", MODE="0660"
-SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="8010", GROUP="dialout", MODE="0660"
+SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="8010", GROUP="plugdev", MODE="0660", RUN+="/bin/sh -c '/usr/bin/setfacl -m g:dialout:rw /dev/%k || :'"
 
 # Programmer in ARM mode
-SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="8012", GROUP="plugdev", MODE="0660"
-SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="8012", GROUP="dialout", MODE="0660"
+SUBSYSTEM=="usb", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="8012", GROUP="plugdev", MODE="0660", RUN+="/bin/sh -c '/usr/bin/setfacl -m g:dialout:rw /dev/%k || :'"
 
 # Programmer in IAP mode
-SUBSYSTEM=="usb", ATTRS{idVendor}=="4348", ATTRS{idProduct}=="55e0", GROUP="plugdev", MODE="0660"
-SUBSYSTEM=="usb", ATTRS{idVendor}=="4348", ATTRS{idProduct}=="55e0", GROUP="dialout", MODE="0660"
+SUBSYSTEM=="usb", ATTRS{idVendor}=="4348", ATTRS{idProduct}=="55e0", GROUP="plugdev", MODE="0660", RUN+="/bin/sh -c '/usr/bin/setfacl -m g:dialout:rw /dev/%k || :'"
 
-SUBSYSTEM=="usb", ATTRS{idVendor}=="303a", ATTRS{idProduct}=="4004", GROUP="plugdev", MODE="0660"
-SUBSYSTEM=="usb", ATTRS{idVendor}=="303a", ATTRS{idProduct}=="4004", GROUP="dialout", MODE="0660"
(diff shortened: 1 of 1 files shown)


cnlohr:
I don't think this is correct. We should not be messing with these things on people's computers.

People can always do

```
usermod -aG plugdev <username>
```

YuukiUmeta-UIAP:
> I don't think this is correct. We should not be messing with these things on people's computers.
> 
> People can always do
> 
> ```
> usermod -aG plugdev <username>
> ```

I understand. I changed to only fixing a single GROUP definition for a single device.

cnlohr:
I don't see any actual changes here.  Are you just adding the comments so people are aware?  I think it's probably OK if you uncomment them in case users are on non-plugdev (and instead dialout) systems.

YuukiUmeta-UIAP:
Because code is confusing, I want to change.

```
DEVICE=="foo", GROUP="plugdev", MODE="0660"
DEVICE=="foo", GROUP="dialout", MODE="0660"
```

This code overwrites foo device node state `root:plugdev` to `root:dialout`. However, at first glance it appears to configure to be accessible to both groups. Additionally, my group was just plugdev with no dialout, then I couldn't access to the device on my Ubuntu 24.04.

My understanding is that a single device connected to a PC has a single group because it has single node state like `owner:group`. And access rights for multiple groups can be set only other ways for example using TAG+="" or ACL.

Sorry my bad at explaining my intentions.

TommyMurphyTM1234:
> Sorry my bad at explaining my intentions.

I don't understand this whole PR at all. 

`plugdev` is the conventional group on many or most popular GNU/Linux distros to allow non `root` access to USB devices. 
I don't understand why you would use the `dialout` group instead (or as well - it's really not clear).

Please explain the perceived issue and proposed solution more clearly. Thanks.

YuukiUmeta-UIAP:
> Please explain the perceived issue and proposed solution more clearly. Thanks.

Sure!

First, I tried to fix to be allow multipl gruop with ACL. because the at first glance, the before code appears to configure to be accessible to both groups.

Before udev code
```
DEVICE=="foo", GROUP="plugdev", MODE="0660"
DEVICE=="foo", GROUP="dialout", MODE="0660"
```

After udev code
```
DEVICE=="foo", GROUP="plugdev", MODE="0660", RUN+="/bin/sh -c '/usr/bin/setfacl -m g:dialout:rw /dev/%k || :'"
```

> I don't think this is correct. We should not be messing with these things on people's computers.

I understood what he mean, and I changed fixing to multipl group to single group.

Before udev code
```
DEVICE=="foo", GROUP="plugdev", MODE="0660"
DEVICE=="foo", GROUP="dialout", MODE="0660"
```

After udev code
```
DEVICE=="foo", GROUP="plugdev", MODE="0660"
#DEVICE=="foo", GROUP="dialout", MODE="0660"
```

Do I need to do more clear? or, Do I have some wrong of premise recognition?

cnlohr:
Putting the # in front makes it so even if a user is part of dialout, they don't get access.  Some Linux distros use dialout.  Some Linux distros use plugdev, I'd like to allow either, or, or both. I don't think this is a good idea?

YuukiUmeta-UIAP:
I agree with you.

But my understanding is that udev doesn't allow multiple GROUPs because a device node is managed by the file system and has a single group like a file permission as `root:root`.

In other words, the before udev code does not works as intended that you want to allow both.
```
DEVICE=="foo", GROUP="plugdev", MODE="0660"
DEVICE=="foo", GROUP="dialout", MODE="0660"
```
The actual internal processing may be different, but I interpret the before udev code as follows:
```
chown root:plugdev /dev/foo
chown root:dialout /dev/foo
```
Therefore, the before udev code can't allow both GROUPs to access. Only `dialout` can access.

To sum up, I want to comment its second line out. Because the before udev code is works without any problems, but it's confusing.

(Allow me to add a little more information. In first commit, I tried to allow multiple access using ACL. But it is danger because complex code.)

maxgerhardt:
Would it be better to change the mode to just 666 so that every user can have acess to it? That's how PIO does it. They also [document this](https://docs.platformio.org/en/stable/core/installation/udev-rules.html#alternative-using-group-membership)

>The alternative permanent solution is to add its own “username” to the “dialout” group, or whichever group name was identified at the preceding step. Typical names are “dialout”, “plugdev” (Debian/Ubuntu, Fedora), or “uucp” (Arch Linux). Adding a user to a group is done by:
>
>`sudo usermod -a -G dialout $USERNAME`


https://github.com/platformio/platformio-core/blob/487a894a713d36acbd9d858267076d0ab957d543/platformio/assets/system/99-platformio-udev.rules#L97-L105

cnlohr:
If we have a new PR that addresses these issues that would be great.  Hopefully what we have now is suitable.

PR #673, opened 2025-08-15
ch5xx: Merge ch5[789]xhw.h files into ch5xxhw.h

biemster:
This is beautiful, I should have done this from the start. The whole ch5xx series is basically the same thing just with random peripherals glued on!

- [x] negative code ! :partying_face: 
- [ ] There is also a fix for the minichlink debug terminal for ch582 (reverted, already went in in #674 )
- [x] The demos all have gotten comments on how they work.
- [x] Also the full ch585 SFR will be added in this package.


biemster:
This is done :tada:.

PR #675, opened 2025-08-18
Fix Arduino-like PU and PD configuration when using CH32V003:

YuukiUmeta-UIAP:
- Extend funPinMode to be able to receive pull-up and pull-down alias
- Enable the alias (configuration mode enumeration) on ch32v003hw.h

Take care if you merge!! I verified only follows:
- No any different binary size on those codes.
  - `funPinMode( PXX, GPIO_Mode_IPU | GPIO_CNF_IN_PUPD )`
  - `funPinMode( PXX, GPIO_Mode_IPD | GPIO_CNF_IN_PUPD )`
  - `funPinMode( PXX, GPIO_CNF_IN_PUPD )`
- Pull-up level works on my CH32V003 with checking a voltage tester.
- Preprocessed code with `riscv64-unknown-elf-gcc -E` is the same as my intent.
- `funPinMode( PXX, GPIO_Mode_IPU )` and `funPinMode( PXX, GPIO_Mode_IPD )` are work fine too.


YuukiUmeta-UIAP:
I added "processing to prevent the `mode` digits from increasing" that I had forgot to write when fair copy.

YuukiUmeta-UIAP:
And I added fixing to prevent `__ASSEMBLER__` interference on rv003usb.

YuukiUmeta-UIAP:
I verified no warning using only `funPinMode( PXX, GPIO_CFGLR_OUT_10Mhz_PP )`.

cnlohr:
I am happy to accept the __ASSEMBLER__ fixes but we can't accept funPinMode getting more complicated, sorry :(

PR #676, opened 2025-08-19
Add v20x ADC example

JackToaster:
Add a simple polling ADC example (much like the v003 version) that works on the ch32V20x. Tested on a V203C6T6 on a NanoCh32v203 dev board and a custom V203 dev board.

### DIFF
--- a/examples_v20x/adc/adc.c
+++ b/examples_v20x/adc/adc.c
@@ -1,0 +1,80 @@
+#include "ch32fun.h"
+#include "ch32v20xhw.h"
+#include <stdio.h>
+
+
+void adc_init( void )
+{
+	// ADCCLK = 12 MHz (APB2 CLK=72MHz, divide by 6)
+
+	RCC->CFGR0 &= ~RCC_ADCPRE;
+	RCC->CFGR0 |= RCC_ADCPRE_DIV6;
+	
+	// // Enable GPIOA and ADC
+	RCC->APB2PCENR |= RCC_APB2Periph_GPIOA | RCC_APB2Periph_ADC1;
+	
+	// // PA7 is analog input chl 7
+	GPIOA->CFGLR &= ~(0xf<<(4*7));	// CNF = 00: Analog, MODE = 00: Input
+	
+	// Reset the ADC to init all regs
+	RCC->APB2PRSTR |= RCC_APB2Periph_ADC1;
+	RCC->APB2PRSTR &= ~RCC_APB2Periph_ADC1;
+	
+	// Set up single conversion on chl 7
+	ADC1->RSQR1 = 0;
+	ADC1->RSQR2 = 0;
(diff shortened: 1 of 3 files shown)


examples_v20x/adc/adc.c
```diff
@@ -0,0 +1,80 @@
+#include "ch32fun.h"
+#include "ch32v20xhw.h"
+#include <stdio.h>
+
+
+void adc_init( void )
+{
```
cnlohr, line 7: Please articulate how this is different from the regular ADC demo in the examples folder.

examples_v20x/adc/adc.c
```diff
@@ -0,0 +1,80 @@
...
+	// ADCCLK = 12 MHz (APB2 CLK=72MHz, divide by 6)
+
+	RCC->CFGR0 &= ~RCC_ADCPRE;
+	RCC->CFGR0 |= RCC_ADCPRE_DIV6;
+	
+	// // Enable GPIOA and ADC
+	RCC->APB2PCENR |= RCC_APB2Periph_GPIOA | RCC_APB2Periph_ADC1;
+	
+	// // PA7 is analog input chl 7
+	GPIOA->CFGLR &= ~(0xf<<(4*7));	// CNF = 00: Analog, MODE = 00: Input
```
cnlohr, line 17: Please use new funConfig functionality for this.

Also the initall function.

cnlohr:
@JackToaster are you up for tweaking your code to my comments?

JackToaster:
@cnlohr Haven't had much time to work on this in the last few weeks, but I recently got scanning conversion working on the '203 as well. I think that would be a better example than the basic polling conversion since the polling ADC example for the '203 is so similar to the '003 version.
I can update the example to use scanning conversion later today (or close this and make a new MR for that if you'd prefer)

cnlohr:
Either which way, just wanted to let you know your help is welcome!

hexaedron:
@cnlohr I can confirm that `funAnalogInit()` does not work properly on 208. The difference is:

```c
// turn on ADC and set rule group to sw trig
ADC1->CTLR2 |= ADC_ADON | ADC_EXTSEL | ADC_EXTTRIG;
```

Without `ADC_EXTTRIG`, conversion lasts forever in `funAnalogRead()`. I have no idea why. If we set `ADC_EXTTRIG` as @JackToaster proposed in his example, it works fine.

hexaedron:
Also, one should remember that ADC can not work if its clock is above 14MHz.

cnlohr:
@hexaedron I would really appreciate progress in this direction, whether it be in the direction of a separate example, or updating funAnalogInit() for support on more platforms.

cnlohr:
I am happy to accept this in the future, but for now it looks like this has comment rotted.

PR #677, opened 2025-08-20
Add GPIOB and USART2 based on CH32V00XRM.PDF

YuukiUmeta-UIAP:
I have NOT confirmed that it works fine.


cnlohr:
I don't think you need to test them.  I've just reviewed this against the datasheet and I confirm that these are correct at least as-written.

PR #678, opened 2025-08-21
fix CH582_CH582 typo in SysTick

biemster:
The SysTick code got a CH582_CH582 typo in #673, this fixes that (and makes uni happy!)


biemster:
the ci is not aging well, jenkins can you try a rebuild?

cnlohr:
Eh that was a good enough try for CI.

But yeah, I am thinking we should move to fewer spawned tasks.

PR #685, opened 2025-08-23
Fix #107 issue in rv003usb

m039:
I've reordered the statements and it looks like it fixes https://github.com/cnlohr/rv003usb/issues/107

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -1633,7 +1633,6 @@ void SystemInit( void )
 //#define BASE_CTLR	(((FUNCONF_HSITRIM) << 3) | HSEBYP | RCC_CSS)	// disable HSI in HSE modes
 
 	// Flash latency settings.
-	// Flash latency settings.
 #if defined(CH32V003)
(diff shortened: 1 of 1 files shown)


monte-monte:
ch32fun is a submodule. You need to make changes to it in ch32fun repo. Also try updating it first, maybe it was fixed there.

m039:
But I need to reference this commit in rv003usb even it is a submodule there, so I need to merge this commit first, isn't it? And I think the problem is more general than only related to rv003usb.

cnlohr:
Well, this should never be an issue because `CH32V00x` should not be defined if `CH32V003` is defined.  But this _does_ make the code more robust so I will accept it.

maxgerhardt:
Is a CH32V003 not a CH32V00x? Then x is not generic. I have `CH32V00x` as the generic identification macro for a CH32V003 as well as literally all other CH32V00{2,3,4,5,6}. Mhm, complicated.

monte-monte:
Sorry I got confused (probably with multiple opened tabs). I thought I was commenting in rv003usb.

cnlohr:
whelp this may may it so it doesn't really matter.

PR #686, opened 2025-08-24
Add mcpy demo for CH570/2 and CH585

monte-monte:
Newly found custom WCH instruction ``mcpy`` that supported in CH570/2 and CH585 chips. It greatly increases memory copying speed. But you need to respect 4 bytes alignment. Also it's probably copies in 4 bytes chunks.


cnlohr:
APPROVE APPROVE APPROVE

PR #688, opened 2025-08-30
Drop dependency on "expr", add Windows Makefile build to CI

maxgerhardt:
* Drops depedency on `$(shell expr '$(PREFIX)-gcc -dumpversion | cut -f1 -d.' \>= 13)` by using Make builtins like `firstword`, `filter` and `if` to achieve the same result
* Surpresses noisy Windows errors when executing `where` on non-installed programms, it would print `INFO: Could not find files for the given pattern(s).` over and over again. (Redirection of stderr to `nul` or `/dev/null` on Linux, leaving stdout intact)
* Adds a native Windows build job that compiles all ch32v003 examples in the `examples/` folder using `make` via `build_scripts/build_all.cmd`
  * uses and thus also tests existing `install_xpack_gcc.ps1` script to install dependencies and setup environment
    * small addition to detect Github runners and add it to the special `$GITHUB_ENV` environment variable too, instead of just `$PATH`  
  * doesn't build any of the other examples folders due to wanting to keep it fast and lean

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,19 @@ jobs:
                 echo "============================"
                 (cd "$dir" && make V=1 -j3 "$(basename "$dir").elf" && riscv64-unknown-elf-size "$(basename "$dir").elf")
             '
+  build-all-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


cnlohr:
Thank you so much for doing this, @maxgerhardt

PR #689, opened 2025-09-01
Add primitive blink support for CH32L103

nolanhergert:
Mostly copied from CH32V203 implementation.

PLL multiplier more than 6 doesn't start. I've confirmed the same board works up to 16X multiplier, so it's likely not a hardware issue.

Have not debugged or looked into much else.


cnlohr:
Please update https://github.com/cnlohr/ch32fun/blob/master/ch32fun/ch32fun.ld

**EDIT** Never mind I see in the code that it's already handled.

ch32fun/ch32fun.h
```diff
@@ -64,7 +64,7 @@
 /******************************************************************************
  * CH32V003 Fun Configs; please define any non-default options in funconfig.h *
 
-#define FUNCONF_USE_PLL 1               // Use built-in 2x PLL
```
cnlohr: In general, please watch for this.  Whitespace changes are OK, but they should be separate commits.

ch32fun/ch32fun.c
```diff
@@ -1827,10 +1827,10 @@ int funAnalogRead( int nAnalogNumber )
 
 	// start sw conversion (auto clears)
 	ADC1->CTLR2 |= ADC_SWSTART;
-
```
cnlohr: Whitespace commits are OK but should be separate commits... All that said I will accept this as is, but for the future.

ch32fun/ch32fun.mk
```diff
@@ -120,7 +120,25 @@ else ifeq ($(findstring CH32X03,$(TARGET_MCU)),CH32X03) # CH32X033, X035
...
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring K8, $(TARGET_MCU_PACKAGE)), K8)
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring C8, $(TARGET_MCU_PACKAGE)), C8)
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring G8, $(TARGET_MCU_PACKAGE)), G8)
+		MCU_PACKAGE:=1
+	endif
+
+	TARGET_MCU_LD:=4
```
cnlohr, line 141: Oh never mind I see you don't need to update the LD file.

minichlink/pgm-wch-linke.c
```diff
@@ -73,7 +73,7 @@ void wch_link_command( libusb_device_handle * devh, const void * command_v, int
 
 	status = libusb_bulk_transfer( devh, 0x81, reply, replymax, transferred, WCHTIMEOUT );
 
-//	printf("wch_link_command reply (%d)", *transferred); for(int i = 0; i< *transferred; printf(" %02x",reply[i++])); printf("\n");
```
cnlohr: Ok, all of the whitespaces here are really annoying.  Next time please don't.

cnlohr:
@biemster or @monte-monte would you like to review?

How do you feel about people making both whitespace and regular changes to code in one place?

biemster:
> @biemster or @monte-monte would you like to review?
> 
> How do you feel about people making both whitespace and regular changes to code in one place?

I never really thought about it, I do it regularly as well actually (although not as extensive as nolan :clap: ). I'm usually only a bit worried that it will mess up the blame, but since I never had an issue caused by this I guess blame is smart enough to disregard whitespace?

For reviews it does not bother me.

ch32fun/ch32fun.h
```diff
@@ -197,6 +197,8 @@
 	#if defined(FUNCONF_USE_PLL) && FUNCONF_USE_PLL
 		#if defined(CH32V10x)
 			#define FUNCONF_PLL_MULTIPLIER 10	// Default: 8 * 10 = 80 MHz
+		#elif defined(CH32L103)
+			#define FUNCONF_PLL_MULTIPLIER 6    // TODO: Why won't 12 run?
```
biemster: Maybe you need a different flash setting for higher speeds?

ch32fun/ch32v20xhw.h
```diff
@@ -8062,7 +8062,7 @@ typedef struct
 	#if !defined(FUNCONF_SYSTEM_CORE_CLOCK)
 		#define PLL_MULTIPLICATION ((uint32_t)0)
 	#else
-		#if defined(CH32V10x) || defined(CH32V20x) || defined(CH32V30x_D8)
+		#if defined(CH32V10x) || defined(CH32V20x) || defined(CH32L103) || defined(CH32V30x_D8)
```
biemster: This shouldn't be needed, it is likely an artifact from when the different chip series were split in different header files. Can you try without it?

ch32fun/ch32fun.h
nolanhergert: Sounds good, I will update those later today/tomorrow.

ch32fun/ch32fun.h
nolanhergert: That would definitely explain it! Let me try later today/tomorrow.

ch32fun/ch32v20xhw.h
nolanhergert: Sounds good, fine with me. I wasn't sure if you wanted them to be consistent still post-split.

ch32fun/ch32fun.h
cnlohr: Oh good point - wait states and all!

cnlohr:
Ok, if you can get this a little bit further along, @nolanhergert and please squash your merges into one commit for this if possible.

ch32fun/ch32fun.h
nolanhergert: Works great, including up to PLL multiplier of 18 (only tested blink though). I will lower it down to the default of 96MHz but hint that overclocking is possible.

nolanhergert:
Reverted all whitespace changes and only made the necessary changes from ch32v20xhw.h -> ch32l103.h, so the diff is easier. Code blinks and runs at 96Mhz.

cnlohr:
This looks good and clean! Merging!

PR #691, opened 2025-09-02
Provide default Handlers for USBFS CDC_TTY

BogdanTheGeek:
This should simplify the setup.
Questions:
 - should I also check if `FUSB_USER_HANDLERS` is set or will it always be set if `FUNCONF_USE_USBPRINTF` is set?
 - should I modify the cdc_tty example or should I add a new minimal usbprintf example?

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -92,8 +92,6 @@ void __libc_init_array(void)
 #include <stdint.h>
 #include <ch32fun.h>
 
-#define WEAK __attribute__((weak))
-
 WEAK int errno;
 
 static int __puts_uart( char *s, int len, void *buf )
--- a/ch32fun/ch32fun.h
+++ b/ch32fun/ch32fun.h
@@ -299,6 +299,9 @@
 
 #endif
 
+#ifndef WEAK
+#define WEAK __attribute__((weak))
+#endif
 
 #ifdef __cplusplus
  extern "C" {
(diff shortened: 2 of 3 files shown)


cnlohr:
looks good

biemster:
> This should simplify the setup. Questions:
> 
> * should I also check if `FUSB_USER_HANDLERS` is set or will it always be set if `FUNCONF_USE_USBPRINTF` is set?
> * should I modify the cdc_tty example or should I add a new minimal usbprintf example?

The `FUSB_USER_HANDLERS` is probably a @monte-monte question, and I'd prefer to modify the already existing example (although I do like the `usbprintf` name more, why did I not think of that)

monte-monte:
@biemster lol, no. I wanted you to decide on it :D
Anyway, this is for a specific situation, so I don't see why not.

biemster:
I just noticed now that the WEAK thing is not going to fly, because `fsusb.h` is `#include`ing `fsusb.c`. This means the compiler will see both, and throw a redefinition error. This is also causing #707 .

TommyMurphyTM1234:
> `fsusb.h` is `#include`ing `fsusb.c`.

Why? Seems like a very odd coding style...

- https://github.com/cnlohr/ch32fun/blob/7a25825af8381f6ad696be0599ce7c985283eb50/examples_x035/usbdevice/fsusb.h#L87

monte-monte:
@TommyMurphyTM1234 it is to don't need to add fusb.c to the makefile.

biemster:
How did this even compile? When I checkout on this hash I get errors.

PR #692, opened 2025-09-03
Changed output stream for non-error fprintf for minilink uploading

Themuse061:
Change output stream from stderr to stdout for non-error fprints so that `> /dev/null` can be used to silence command
example

for command
`$minichlink -w binary.bin flash -b > /dev/null`

changes output from
image
to
image

while still printing errors
image
(example with LinkE with no MCU connected)


BogdanTheGeek:
makes sense, odd that those were going into stderr.

cnlohr:
We were trying to direct as much as possible to stderr so that we can do:

```
minichlink -T > file_transferred_from_chip.txt
```

We don't have to but I have a strong preference to be able to continue to use it as a proper pipe.

cnlohr:
That said if @monte-monte or someone wanted to make it less chatty, that would be welcome.

monte-monte:
I was using ``fprintf`` to ``stderr`` as I saw it was already used in other places, and I noticed that fprintf is flushed immediately, while printf can be printed out of order, so if there is any better way of doing it I will welcome the change. Can't you just forward stderr to ``/dev/null`` also?

Themuse061:
I wanted to be able to mute normal output while keeping error messages to keep my output console clean while still keeping error messages. Would it be possible to implement --silence and  --silence_all as an option?

TommyMurphyTM1234:
> Can't you just forward stderr to `/dev/null` also?

You mean this perhaps?
```
minichlink -T 2>&1 > file_transferred_from_chip.txt
```
(There's some equivalent for Windows Command or Power Shells but I'm too lazy to look it up... ;-)

Themuse061:
nope, i wanted to output error messages to console while silencing rest of output by some sort of --silent option.
so you could use
`minichlink -T > file_transferred_from_chip.txt` to save to file
and use
'minichlink -w binary.bin flash -b --silent' when you want to upload without any unnecessary stuff printed to console

problem with redirecting stderr ( so 2>&1 ) is that when some error occurs, it wont get printed to console (cuz it got redirected) and you will only see that make finished with an error

I would do that, but i don't know/understand how the options are passed from command line to C

TommyMurphyTM1234:
> nope

I was replying to @monte-monte's [question](https://github.com/cnlohr/ch32fun/pull/692#issuecomment-3259968060).

cnlohr:
typically that is more of a --quiet which just makes things much less chatty. That I'd be happy to support. (Though I don't have time at the moment to write it myself)

Themuse061:
Okay, I've implemented --quiet. Tested on windows but also should work on linux. Right now I don't have acces to my wchlink so i might missed some places to replace stdout with my output stream

to test you need to run 
`$minichlink -w binary.bin flash -b --quiet`

minichlink/minichlink.c
```diff
@@ -149,6 +152,21 @@ int main( int argc, char ** argv )
 		}
 	}
 
+	normal_stream = stderr;
+	
+	// Scan for --quiet
+	for( i = 0; i < argc; i++ )
+	{
+		char * v = argv[i];
```
cnlohr, line 160: weird to add a new place here, but it needs to be done.

minichlink/minichlink.c
```diff
@@ -149,6 +152,21 @@ int main( int argc, char ** argv )
 		}
 	}
 
+	normal_stream = stderr;
```
cnlohr, line 155: Please do not set this here. Please set it in the very very beginning of `main` or at file scope, as the default initializer.

cnlohr:
Very close, but this will potentially crash if done within a DLL or library.

Themuse061:
I'm not really sure how to make tests for that, so I'll keep this change on my fork for now

PR #693, opened 2025-09-03
Access "undocumented" flash on v20x

BogdanTheGeek:
The flash memory for all v20x is actually a 256k external spi die bonded to the chip. The flash mentioned in the DS is actually the size of the cache that allows no wait access, but the rest of the chip can still be used.
This is still WIP, but I was able to link code into the slower flash and run from it.
Here are some benchmarks that i ran for read only:
```
Slow Flash read 4096 bytes in 35592 ticks
Fast Flash read 4096 bytes in 5634 ticks
Slow flash read is 631% slower or 15% the speed of fast flash
```
and executing from it:
```
Slow@08020000: took 3512525 ticks, result 3205032704
Fast@0000027A: took 137504 ticks, result 3205032704
Slow flash read is 2554% slower or 3% the speed of fast flash
```

### DIFF
--- a/ch32fun/ch32fun.ld
+++ b/ch32fun/ch32fun.ld
@@ -2,10 +2,10 @@ ENTRY( InterruptVector )
 
 MEMORY
 {
-#if TARGET_MCU_LD == 0
+#if TARGET_MCU_LD == 0 /* v00x */
 	FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 16K
 	RAM (xrw)  : ORIGIN = 0x20000000, LENGTH = 2K
-#elif TARGET_MCU_LD == 1
+#elif TARGET_MCU_LD == 1 /* v10x */
 	#if MCU_PACKAGE == 1
 		FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 64K
 		RAM (xrw) : ORIGIN = 0x20000000, LENGTH = 20K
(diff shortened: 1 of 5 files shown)


BogdanTheGeek:
I've realised that changing the linker to place the FLASH section at 0x08000000 and telling minichlink to flash there would simplify a few things as the binary would be contiguous. However, it does mean it will have to flash a bunch of 0s if you put something in the storage section. Not sure which one would be preferred.

BogdanTheGeek:
Ok so there is on small problem changing the origin, the entry point jump cannot address more than 20bit, so it needs 2 instructions to jump that far. If the jump is relative to 0x00, gcc is too clever for itself and uses a bunch of pc relative jumps which break.
With the following changes, it boots fine regardless of where you compile the code:
```diff
diff --git a/ch32fun/ch32fun.ld b/ch32fun/ch32fun.ld
index c872151..ea6e9b5 100644
--- a/ch32fun/ch32fun.ld
+++ b/ch32fun/ch32fun.ld
@@ -17,7 +17,7 @@ MEMORY
 	#endif
 #elif TARGET_MCU_LD == 2 /* v20x */
 	#if MCU_PACKAGE == 1
-		FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 64K
+		FLASH (rx) : ORIGIN = 0x08000000, LENGTH = 64K
 		EXT (rx) : ORIGIN = 0x08010000, LENGTH = 192K
 		RAM (xrw) : ORIGIN = 0x20000000, LENGTH = 20K
 	#elif MCU_PACKAGE == 2
diff --git a/ch32fun/ch32v20xhw.h b/ch32fun/ch32v20xhw.h
index 451f258..9a9f80d 100644
--- a/ch32fun/ch32v20xhw.h
+++ b/ch32fun/ch32v20xhw.h
@@ -98,8 +98,8 @@ typedef enum IRQn
 	.balign  2\n\
 	.option   push;\n\
 	.option   norvc;\n\
-	j handle_reset\n\
-	.word   0 \n\
+	lui   a0, %hi(handle_reset) \n\
+	jalr a0, %lo(handle_reset) \n\
 	.word   NMI_Handler                /* NMI */ \n\
 	.word   HardFault_Handler          /* Hard Fault */ \n\
 	.word   0 \n\
```

I think i am starting to prefer the idea of just having a separate partition, but I am happy to report that its possible to compile into 0x0800 0000 with no performance penalty.

BogdanTheGeek:
@cnlohr any opinion on this?

cnlohr:
Sorry I saw it was draft. 

I really like this ... I don't know why we wouldn't do that.

PR #696, opened 2025-09-07
ssd1306: 128x32 OLED - Proper SSD1306_SETCOMPINS Configuration

SadaleNet:
I was writing my own ssd1306 OLED driver for 128x32 ssd1306 display. I used your ssd1306 library as reference. I couldn't help but noticed that there's an awkward "expand" array here in the source code: https://github.com/cnlohr/ch32fun/blob/0e72a0ad1c51b5d1a68f27cb6edc4f7d1cd9face/extralibs/ssd1306.h#L195

Apparently it's used as a workaround that 128x32 OLED could only display pixels on odd rows.

However, after experimenting for a while, I've found way to make 128x32 OLED to display all the rows. It seems that the display odd-row only behavior was caused by improper configuration of the `SSD1306_SETCOMPINS` variable. The fix is to use `SSD1306_SETCOMPINS 0x22` instead of `SSD1306_SETCOMPINS 0x12` in the initialization sequence.

Original pixel addressing, each spreadsheet cell is a pixel:
image

Pixel addressing with proposed change:
image

Considered the popularity of this library, I find it odd that no one had noticed that before. I wonder if I'm missing anything.

I've only tested the `i2c_oled` example code with the 128x32 OLED unit that I've got on my hand. Please help carefully review if this change would have further impact before merging.

### DIFF
--- a/extralibs/ssd1306.h
+++ b/extralibs/ssd1306.h
@@ -156,7 +156,11 @@ const uint8_t ssd1306_init_array[] =
     SSD1306_SEGREMAP | 0x1,                // 0xA0 | bit
     SSD1306_COMSCANDEC,
     SSD1306_SETCOMPINS,                    // 0xDA
+#if defined(SSD1306_FULLUSE)
 	0x12,                                  //
+#else
+	0x22,                                  //
+#endif
     SSD1306_SETCONTRAST,                   // 0x81
 #ifdef SSD1306_72X40
 	0xAF,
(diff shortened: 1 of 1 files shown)


emeb:
Nice work! Yes that expansion process is just as you inferred. If this gets rid of it then I'm glad to see  it go away.

cnlohr:
Will merge when workflows are done.  I take @emeb as an approval.

emeb:
I've tried this fix and it works on my test OLEDs. Thanks for figuring that out.

PR #699, opened 2025-09-10
Improve i2c_slave documentation, add remap examples

binary1230:
I was banging my head against the wall on this getting it working for the first time, decided to add some sample code for the next person.

This adds some examples on how to do the I2C1 remapping in the AFIO_PCFR1 register, using the latest symbols in Ch32fun [which are named differently than the CH32V003 manual].  I have only tested the 2nd example, though it's pretty simple.

Also just made it a little more obvious you can't put I2C on any random pins, you need to pick specific ones.


cnlohr:
This is great.  Thanks.

PR #702, opened 2025-09-14
Fix for serial port reads on Linux.

gatk555:
I have been trying to use minichlink with an Arduino on Linux.  No luck so far, but one early problem was that the returned register data was nonsense.  Debugging showed an OS issue with the serial port: when asking for a 4 - byte read, less than 4 may be returned.  The documentation (man 3 termios) suggests that there is no good way to block for the requested bytes, unless resetting the port characteristics (VMIN) for each change in read length.  Repeating the read seems less ugly, and harmless if any kernels actually behave as the previous code expects.  (I used Debian 12, kernel 6.1.0-38.)

### DIFF
--- a/minichlink/serial_dev.c
+++ b/minichlink/serial_dev.c
@@ -115,7 +115,19 @@ int serial_dev_read(serial_dev_t *dev, void* data, size_t len) {
 	}
 	return (int) dwBytesRead;
 #else
-	return read(dev->fd, data, len);
+	int got, count = 0;
+
+	/* On Linux, raw mode only blocks on the first character: man 3 termios. */
+
+	while (count < len) {
+		got = read(dev->fd, (char *)data + count, len - count);
+		if (got < 0)
+			return got;
+		count += got;
+		if (count < len)
+			usleep((100000 / dev->baud) * (len - count));
+	}
+	return count;
 #endif
 }
 


cnlohr:
Any chance I could get @maxgerhardt and/or @monte-monte to review?

cnlohr:
(I just don't know anything about this driver)

monte-monte:
I don't know either, sorry.

maxgerhardt:
I approve of this. The users of this API only ever check for negative values but not whether it's a positive value matching the number of requested bytes.

https://github.com/cnlohr/ch32fun/blob/7a25825af8381f6ad696be0599ce7c985283eb50/minichlink/ardulink.c#L52-L56

So turning this into a blocking "read until specified number of bytes has been read" seems needed. I just wonder if the Windows path needs the same treatment. Curiously when I tested Ardulink uploading in both Windows (10) and Ubuntu (22 or 24?) in a VM, it did **work**, I did not need that.

maxgerhardt:
It looks to me like we still need it on Windows. So the function 
```c
int serial_dev_read(serial_dev_t *dev, void* data, size_t len) {
#ifdef IS_WINDOWS
    size_t got = 0;
    DWORD dwRead = 0;

    while (got < len) {
        if (!ReadFile(dev->handle, (char*)data + got, (DWORD)(len - got), &dwRead, NULL)) {
            return -1;  // error
        }
        if (dwRead == 0) {
            // This means timeout or device closed. Since timeouts are set to infinite,
            // this should normally only happen on disconnect.
            return -1;
        }
        got += dwRead;
    }
    return (int)got;
#else
    int got, count = 0;

    /* On Linux, raw mode only blocks on the first character: man 3 termios. */
    while (count < len) {
        got = read(dev->fd, (char *)data + count, len - count);
        if (got < 0)
            return got;
        count += got;
        if (count < len)
            usleep((100000 / dev->baud) * (len - count));
    }
    return count;
#endif
}
```
should maybe be used. I guess you put in the `usleep()` to prevent a spinning in the loop for each individual byte and instead try to estimate the physical transmission time of the remaining bytes? Does this improve CPU usage significantly? It does increase latency.

cnlohr:
@maxgerhardt would you prefer to accept this then you do the work on top or would you rather do this as a separate PR?

gatk555:
On Windows: I assumed that ReadFile() will return the bytes requested.  The MSDN page supports that, but is fairly vague.  I could add the code to the PR, but it would be untested.

usleep(): intended to prevent a tight loop, but I did not measure.  The delay is the expected transit time of the remaining data, so should not add significant latency.

If this PR is correct, how did Ardulink ever work?  That seems a good question. My guess is that it works with the orginal Arduino Uno, but I have been using an Arduino Every board that implements serial/USB in a secondary microcontroller, itself running an Arduino program.  That may explain different behaviour.

PR #703, opened 2025-09-15
fix: ch32x00xhw AFIOEN bit

kuwoyuki:
2^0 = 1 :P


BogdanTheGeek:
checked the RM and you are right, seems to have slipped though.

cnlohr:
Thanks for catching this.

cnlohr:
Looks like all the CI is passing, so merge time it is!

PR #706, opened 2025-09-23
Only do USB errata workaround for OUT requests

### LINKED ISSUE #701: USB CDC: `USBFS_errata_dont_send_endpoint_in_window` stays set after `GET_DESCRIPTOR`
Hi!
### Problem
When using the USB CDC example on **CH32X035**, data cannot be sent to the host PC WIN11 (though receiving works).  

After debugging, I found that the flag `USBFS_errata_dont_send_endpoint_in_window` is always set to `1` and never cleared after a `GET_DESCRIPTOR` (`0x06`) request.  

### Root cause
https://github.com/cnlohr/ch32fun/blob/ab05c9409f21c6414cc4ba6215592c11c6ec84cd/extralibs/fsusb.c#L206C1-L216C8
In `HandleDataOut`, this flag is set unconditionally whenever `ctx->USBFS_SetupReqLen == 0`.  
But `GET_DESCRIPTOR` is an **IN request**, so the flag should not be set in this case.  
As a result, all further calls to `USBFS_SendEndpoint` and `USBFS_SendEndpointNEW` are blocked.  

### Fix
Only set the flag for **OUT requests** (like `SET_LINE_CODING`), not for IN requests.  

### Patch
```c
							// Only set this flag for OUT requests (e.g. SET_LINE_CODING).
							// For IN requests (e.g. GET_DESCRIPTOR), don't set it.
							if ( ( ctx->USBFS_SetupReqType & USB_REQ_TYP_IN ) == 0 )
							{
								ctx->USBFS_errata_dont_send_endpoint_in_window = 1;
							}
```

### Result

* `GET_DESCRIPTOR` now works correctly
* USB CDC can send data to the host as expected
* OUT requests (like `SET_LINE_CODING`) are still handled properly

maxgerhardt:
Fixes #701. PR created for discussion.

### DIFF
--- a/extralibs/fsusb.c
+++ b/extralibs/fsusb.c
@@ -210,7 +210,12 @@ void USBFS_IRQHandler()
 							if( ctx->USBFS_SetupReqCode == HID_SET_REPORT )
 								HandleHidUserReportOutComplete( ctx );
 #endif
-							ctx->USBFS_errata_dont_send_endpoint_in_window = 1;
+							// Only set this flag for OUT requests (e.g. SET_LINE_CODING).
+							// For IN requests (e.g. GET_DESCRIPTOR), don't set it.
+							if ( ( ctx->USBFS_SetupReqType & USB_REQ_TYP_IN ) == 0 )
+							{
+								ctx->USBFS_errata_dont_send_endpoint_in_window = 1;
+							}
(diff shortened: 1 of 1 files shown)


cnlohr:
Thanks!

PR #708, opened 2025-09-26
Make funAnalogInit() work on ch32v208

hexaedron:
Based on #676


cnlohr:
I was expecting a `#if` based on the platform.

Will the exttrig bit not cause issues on other platforms?

hexaedron:
@cnlohr You're right. It is better to restrict my change for only one platform - ch32v20x. I've tested it on 208, but according to #676, it should work fine on 203 too.
I suspect it to work on 30x, but i have no 30x dev boards to test for now.

cnlohr:
Sounds good.

PR #709, opened 2025-09-26
CI: also build test examples_usb/USBFS

biemster:
Currently the build test only finds Makefiles in demos directly in the `examples_*` directories, but the `examples_usb` has one extra level: `USBFS`. This change increases the search depth by one to also catch the USB demos.


biemster:
This will pass the CI when #711 (or something similar) is merged.

cnlohr:
I'll just merge it even though I don't like the weak thing

PR #710, opened 2025-09-26
added 74hc595 example.

unicab369:


examples/hc595_demo/hc595_demo.c
```diff
@@ -0,0 +1,182 @@
...
+
+void SPI_Configure() {
+	// reset control register
+	SPI1->CTLR1 = 0;
+
+	// Enable GPIO Port C and SPI peripheral
+	RCC->APB2PCENR |= RCC_APB2Periph_GPIOC | RCC_APB2Periph_SPI1;
+
+	// PC5 is SCLK
+	GPIOC->CFGLR &= ~(0xf << (4*5));
```
cnlohr: There is no value in writing this out.  Please use `funPinMode(  PC5, GPIO_Speed_50MHz | GPIO_CNF_OUT_PP_AF );` and the same for the ones below.

cnlohr:
Just that one request.  Everything else looks OK.

examples/hc595_demo/hc595_demo.c
unicab369: updated.

Issue #707, opened 2025-09-24
CH5xx USB example compilation error


Randomblock1:
Currently both USB examples fail to compile for CH5xx. The TTY example is just some compaints about redefinitions, but UART example errors indicate there's some bigger issues. I'm trying to get this to work for CH570D, but it also gives some errors with CH592. I unfortunately don't have enough time to fix it myself, but I can at least tell you there's a missing semicolon on line 50 of uart.h.

On a related note, is there a way to get CH570 support working with PlatformIO? AFAIK it's not in there yet because a new version of ch32fun hasn't been released yet or something. I tried adding a custom board but it didn't seem to want to work with ch32fun framework.

biemster:
Thanks for pointing this out! I just tried and am seeing the same, I'm surprised the CI did not catch this @cnlohr ?

cnlohr:
Thanks for resolving this, @biemster

biemster:
this is not resolved!

biemster:
the USB UART example is indeed not ready for CH5xx yet, I don't think there are any DMA examples there yet.

PR #711, opened 2025-09-27
remove Handlers from usb_tty demo

biemster:
As discovered in #707, the examples_usb demos are not build tested by CI, and the usb_tty example does not compile at all. This change will fix it. Then #709 will succeed as well and should be merged too.

The usb_uart demo also does not compile for ch5xx, but that's because it is not ready for that yet. I will not touch that for now.


cnlohr:
I believe we need the WEAK flags...  That way we can override the functions if need be.

biemster:
WEAK does not work here, as the C code is `#include`d in the header. So the compiler sees them both, and throws a redefinition error. (https://github.com/cnlohr/ch32fun/pull/691#issuecomment-3327665330)

BogdanTheGeek:
> WEAK does not work here, as the C code is `#include`d in the header. So the compiler sees them both, and throws a redefinition error. ([#691 (comment)](https://github.com/cnlohr/ch32fun/pull/691#issuecomment-3327665330))

You are not meant to include fsusb.h more than once, it was designed that way. If you want that feature, add a define guard like:
```c
#ifndef FSUSB_IMPL
#define FSUSB_IMPL 1
#endif

#if FSUSB_IMPL
#include "fsubs.c"
#endif
```

BogdanTheGeek:
ok, I can see what went wrong here, yeah i'm not sure how I didn't spot this when I was working on it, FSUSB_IMPL is needed or just fix the examples to include 'fsusb.c' in the main file, instead of it being included by the header.

biemster:
> ok, I can see what went wrong here, yeah i'm not sure how I didn't spot this when I was working on it, FSUSB_IMPL is needed or just fix the examples to include 'fsusb.c' in the main file, instead of it being included by the header.

If you include `fsusb.c` in the main file, you still can't use WEAK since the problem is that the compiler sees both definitions in the same run right? Ideally `fsusb.c` is compiled parallel to `main.c` (`gcc main.c fsusb.c`, I don't know what's the word)

cnlohr:
Even though I don't like the specific use of weak here, it is blocking #709 - so I will merge this.

PR #713, opened 2025-09-30
enable high_code RAM area for ch58x

biemster:
This was already working for ch59x, now also for ch58x.


biemster:
i just remembered the user stack should go there in the linker script too, I'll commit that here tomorrow

biemster:
ok it's fully done now, please merge? :pleading_face:

PR #716, opened 2025-10-03
SPI chain example

monte-monte:
This examples shows how to setup both SPI devices on the same device and use it combined to make a chained network of equal nodes.


monte-monte:
Currently I've added everything I wanted. So if there is no specific objections...

PR #718, opened 2025-10-15
Add USBHS lib and example

monte-monte:
Adding a *universal* lib for USBHS peripheral and first demo example.

Fixed some minir mistakes in USBFS lib and updated naming of config defines. Logic is that all config defines that are settable in ``usb_confi.h`` are prefixed with ``FUSB``, to indicate where it originates from.

Reverted timer value in unbrick function in minichlink.


cnlohr:
AMAZING! Thank you @monte-monte

PR #720, opened 2025-10-17
Fix CI building wrong projects

maxgerhardt:
Fixes cases where we tried to build testbench Makefiles intended for Linux and not a firmware.

Also makes `make` errors show up in log. (Though they may be further up in the log due to parallelism.)


BogdanTheGeek (approved):
confirmed working

PR #724, opened 2025-10-25
fix Systick typo.

unicab369:
correct typo.


cnlohr:
This should have been done long ago.  Thank you for doing this.

maxgerhardt:
No 

```cpp
#define SysTicK_IRQn  SysTick_IRQn
```
to keep backwards compatibility? This is a breaking change otherwise to all existing ch32fun firmwares that used the previously correct name.

binary1230:
came here to say ... OH! THIS DID CHANGE. I thought I was going crazy for a moment : )

I'm kinda +1 on the #define for backwards compatibility

PR #727, opened 2025-10-27
added adc examples for ch5xx

unicab369:


biemster:
This is great! This was the biggest missing basic feature for ch5xx. I'm going to test it on all the ch5xx I have, and update the table in `examples_ch5xx/README.md` accordingly. (this should not block this merge though). There is also a macro to indicate analog is not available on ch5xx (`FUNCONFIG_INIT_ANALOG`), that can be abolished when this goes in. I'll remove that in the PR that updates the README. Great job @unicab369 and keep them coming!

cnlohr:
This all looks great.

PR #730, opened 2025-10-28
Ch5xx UART send receive

unicab369:


cnlohr:
Using millis is an antipattern.  Please do not use it in exmaples.  We need people to move to the pattern of:

(int32_t)(SysTick - LastEvent) > 3000000 

and similar functionality.  It allows everything to happen without systick interrupts. 

Otherwise people are forced to use a systick IRQ when none is needed.

examples_ch5xx/uart_send_receive/systick_irq.h
```diff
@@ -0,0 +1,79 @@
+/*
```
cnlohr: This is a longer discussion, but I do not think we want to encourage the use of this -- that it is more of an antipattern and we should be seeing if we can get people to use elapsed events instead of needing the systick IRQ.

biemster:
I think this is a very useful demo. If you don't have the time to remove the systick irq @unicab369 , I can take over at some point if git allows me to commit here.

examples_ch5xx/uart_send_receive/systick_irq.h
unicab369: I'm just doing 1 second counter reset for now.

cnlohr: I understand but I'm saying this is a poor pattern for something like this, you should not mimic it.  It's just a reference to show people how they could do it if they really wanted.

Please change your code over to using SysTick->CNT... the way you're using it, you don't need this.  You can say.

```c
uint32_t nextPeriod = SysTick->CNT;

...later

if( (uint32_t)(SysTick->CNT - nextPeriod) > (FUNCONF_SYSTEM_CORE_CLOCK/1000)  )
{
   nextPeriod += (FUNCONF_SYSTEM_CORE_CLOCK/1000);
   // Do other stuff after every 1ms.
}
```

cnlohr: Would it help if we had macros for things like

ELAPSED( SysTick->CNT, nextPeriod ) > ...

PR #732, opened 2025-10-29
RFC: Flag parsing

BogdanTheGeek:
DO NOT MERGE!!!!!
Draft attempt at a new flag parsing methodology. Currently it has the following known limitations:
 - no combined args: -3tb
 - order is not retained (-a -b == -b -a)
 - not all flags have been implemented with new method, fallbacks to old method

Things gained:
 - aliases: -help | -h
 - usage() is updated automatically (there were quite a few flags missing)
 - variables as flags
 - defaults

### DIFF
--- a/minichlink/flag.h
+++ b/minichlink/flag.h
@@ -1,0 +1,1010 @@
+// clang-format off
+// flag.h -- v1.7.1 -- command-line flag parsing
+//
+//   Inspired by Go's flag module: https://pkg.go.dev/flag
+//
+// # Macros API
+//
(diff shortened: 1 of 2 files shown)


cnlohr:
I do have a strong preference for maintaining order, because sometimes I do personally sequence things, especially surrounding unbricking, etc.  I.e. applying power doing X removing power.

Also, I really wish we had long-arguments.

BogdanTheGeek:
We can define multiple arguments that point to the same variable, like
-h/-help.

cnlohr:
@monte-monte  @BogdanTheGeek @dwillmore is there any hope for us to fix flags in minichlink?

I think we have to have --long-parameters

cnlohr:
This branch is no longer compatible btw

PR #733, opened 2025-10-31
CH571/CH573 DMA Segment and Example

iracigt:
This PR adds a new `.dma` segment on the CH571/CH573 and a SPI TX example that uses it. The DMA on the CH571/3 only works for RAM addresses above 0x20004000. This means the lower 2k of RAM from  0x20003800 -  0x20003FFF cannot be used for DMA buffers. A new `.dma` segment in the linker script is guaranteed to fall in the valid region. The startup code is updated to initialize the segment. This is also a blocker for radio support on the CH571/3, which requires DMA.

The included SPI TX demo has been tested on the CH573F. I think it should work on the other CH5xx chips, but I don't have any to test with. Feedback there is appreciated.

For discussion: does it make more sense for this to be an initialized or uninitialized segment? 

Right now it's initialized and I use that in the demo to populate the buffer. I can see an argument that it should be uninitialized because DMA buffers are frequently large and we don't want to bloat the binary image in flash.

This patch only applies to the CH571/3 which have a minimum of 192k of flash vs 18k of RAM. It may be unlikely anyone finds the extra flash usage to be an issue.

### DIFF
--- a/ch32fun/ch32fun.ld
+++ b/ch32fun/ch32fun.ld
@@ -269,19 +269,19 @@ SECTIONS
 		The high 18 bits are assumed to be 0x20004000,
 		yet RAM spans 0x20003800 - 0x20007FFF (18 kB)
 		Thus the lower 2K of RAM are not accessible to the DMA peripherals
-		Here we create a special .dma segement guaranteed to be above 0x20004000.
+		Here we create a special .dma_safe segement guaranteed to be above 0x20004000.
 		To avoid unnecessary padding, this should be the last segment.
 	*/
 
 	/* Force this to be the higher of 0x20004000 and end of previous */
 	/* Mark it NOLOAD so it doesn't occupy any flash space */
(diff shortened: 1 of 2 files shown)


iracigt:
After some brief discussion, I decided to make this section uninitialized. It now occupies no flash space, but designated initializers will be ignored. Any initialization should be done explicitly with `memset` or `memcpy`. For receive buffers, this usually isn't necessary.

cnlohr:
@iracigt do you need to re-base because I just merged #733?

iracigt:
Yes, rebased. I would appreciate it if someone could make sure this didn't break #735. Note there are a couple issues with the way #735 did this that I fixed in this PR. This version fixes .end and avoids allocating `.dma_safe` in flash.

iracigt:
I just tested `examples_usb/USBFS/usbfs_cdc_tty` on my CH573 and it seems to be working just fine.

PR #734, opened 2025-11-02
Ch5xx lowpower sensor

unicab369:


examples_ch5xx/iSLER_sensors_gateway/iSLER_sensors_gateway.c
```diff
@@ -0,0 +1,50 @@
+// Listen to messages from iSLER_sensor_lowpower_node
+
+#include "ch32fun.h"
+#include <stdio.h>
+#include "../iSLER_sensor_lowpower_node/ch5xx_Mess.h"
```
biemster, line 5: Wouldn't it be better to just copy this header here? I'd prefer the examples to be self contained, but I'm open for discussion.

biemster:
This looks fun!

cnlohr:
This is very cool, but I feel like there needs to be a pretty serious reorganization.  
1. Please don't put things in a _Mess file.  If you are making a new iSLER layer, please turn it into a proper iSLER layer instead of throwing some structs in a _mess header file.
2. I don't know how I feel about this as being in the main ch32fun repo. I do lots of ch32fun projects, many of which are their own repos.  It feels like this needs to be explored and bake in a separate repo.

cnlohr:
For the time being I am going to close this PR until there's a significant change. 

I reallyyyy think this should go into its own repo at least for a while.

PR #735, opened 2025-11-03
Add HSI support to USBHS on CH32V30x

monte-monte:
Actively tune HSITRIM by measuring SOF packet timing allows to use HSI as clock source for USBHS on CH32V30x. Without trim it will most probably fail on bigger bulk transfers.

### DIFF
--- a/ch32fun/ch32fun.ld
+++ b/ch32fun/ch32fun.ld
@@ -75,12 +75,13 @@ MEMORY
 	/* CH582/3/4/5 */
 	#if MCU_PACKAGE == 2 || MCU_PACKAGE == 3
 		FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 448K
-		RAM (xrw) : ORIGIN = 0x20000000, LENGTH = 30K
-		HIGHRAM (xrw) : ORIGIN = 0x20007800, LENGTH = 2K
-	#elif MCU_PACKAGE == 4 || MCU_PACKAGE == 5
+		RAM (xrw) : ORIGIN = 0x20000000, LENGTH = 32K
+	#elif MCU_PACKAGE == 4
 		FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 448K
 		RAM (xrw) : ORIGIN = 0x20000000, LENGTH = 96K
-		HIGHRAM (xrw) : ORIGIN = 0x20018000, LENGTH = 32K
+	#elif MCU_PACKAGE == 5
+		FLASH (rx) : ORIGIN = 0x00000000, LENGTH = 448K
+		RAM (xrw) : ORIGIN = 0x20000000, LENGTH = 128K
 	#else
 		#error "Unknown MCU package"
 	#endif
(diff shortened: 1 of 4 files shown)


monte-monte:
ooops...I forgot about this PR, it put all together...

monte-monte:
I've modified ``ch32fun.mk`` to set ``CH5xx`` define, to be able to check for it in ``funconfig.h``. Because the define that biemster added previously couldn't be accessed at funconfig preprocessing stage.

examples_usb/USBFS/usbfs_cdc_tty/usb_config.h
```diff
@@ -5,9 +5,9 @@
 #include "ch32fun.h"
 
 #define FUSB_CONFIG_EPS       4 // Include EP0 in this count
-#define FUSB_EP1_MODE         1 // OUT
-#define FUSB_EP2_MODE        -1 // IN
-#define FUSB_EP3_MODE         1 // OUT
+#define FUSB_EP1_MODE         1 // TX
```
cnlohr: Do not use TX/RX.  I know IN/OUT is confusing, but it is what everyone and the spec uses.  If you want to fix this, say OUT (computer->device)

examples_usb/USBFS/usbfs_device/testtop.hidapi/os_generic.h
```diff
@@ -0,0 +1,521 @@
+#ifndef _OS_GENERIC_H
```
cnlohr, line 1: Can you just reference this file from minichlink instead of here?

examples_usb/USBHS/usbhs_device/usb_config.h
```diff
@@ -5,11 +5,11 @@
 #include "ch32fun.h"
 
 #define FUSB_CONFIG_EPS       6 // Include EP0 in this count
-#define FUSB_EP1_MODE         1 // OUT
-#define FUSB_EP2_MODE         1 // OUT
-#define FUSB_EP3_MODE         1 // OUT
-#define FUSB_EP4_MODE         1 // OUT
-#define FUSB_EP5_MODE        -1 // IN
+#define FUSB_EP1_MODE         1 // TX
```
cnlohr: Also, here, please do not use other nomenclature.  USB endpoints are OUT and IN, not TX and RX.  But explaining what that means is helpful.

monte-monte:
After discussion with biemster we decided to make linker script in line with EVT in part of ``HIGHRAM``. Now separate memory region is gone and section ``.highcode`` is put in the beginning of RAM. Surprisingly it improved performance of USBHS code on CH585 significantly.
Added a special ``.dma_safe`` section for CH571/3 only to fix DMA on this chip. Its DMAs can't access anything below ``0x20004000``. After this change USBFS is working on this chip as well as on any other.

examples_usb/USBFS/usbfs_device/testtop.hidapi/os_generic.h
monte-monte: I'm not sure what you mean. Put ``os_generic.h`` into minichlink folder and point every testtop there?
Because currently there is no ``os_generic.h`` in there.

iracigt:
See #733 for a version of the dma section that does not occupy flash space.

PR #741, opened 2025-11-05
added sleep test.

unicab369:


ch32fun/ch32x03xhw.h
```diff
@@ -277,6 +277,14 @@ typedef struct
 	__IO uint32_t RES3;
 } ESG_TypeDef;
 
+typedef struct
+{
+    __IO uint32_t CSR;
```
cnlohr: uniiiiiiiiiiiiiii tabs

examples_x035/sleep_modes/sleep_modes.c
```diff
@@ -0,0 +1,103 @@
+//# TODO: Figure out why it consumes 12mA in shutdown mode.
```
cnlohr, line 1: Is this still true?

Also, please add details of everything else you are seeing around the chip and how to configure the wiring.

unicab369: ye, I'm not planning on using this chip for low power, so I don't care very much about it right now. But still 12mA is quite high.

examples_x035/sleep_modes/sleep_modes.c
```diff
@@ -4,8 +4,19 @@
 #include <stdio.h>
 #include "debug_utilities.h"
 
-// use defines to make more meaningful names for our GPIO pins
-#define PIN_1 PA9
+// CH32X033F8P6 wirings
```
cnlohr: I don't understand what you are communicating with this table? Why are there so many pins listed?

unicab369: just pins mapping of the Ch32x03xF8P6 package showing where to connect GND, VCC, DIO, and DCK. I think many newbies will be using this package since it's easy to solder. It's just nice to have at least 1 reference of the pin mapping somewhere.

cnlohr: I don't think this demo is the place to explain that.  Perhaps somewhere else, like a readme in the main examples_x035 folder.   

That said - you do need to explain what PA9 is and/or how it's used.

cnlohr:
If this example is stuck at 12mA, I do not want to accept it.  It doesn't show people how to properly do sleep mode.

Our demos have to be very high quality, because they are the basis for others, and they will bring their own jank.  If we start with jank they will be unable to build something good.

cnlohr:
Until you get sleep working good, I'm going to close this for now.

PR #743, opened 2025-11-06
Ch32x03x touch key test

unicab369:


extralibs/register_debug_utilities.h
```diff
@@ -0,0 +1,59 @@
+//! ####################################
+//! PRINT BITS
+//! ####################################
+
+// last 2 args are optional: separator string (default "|"), divider length (default 8)
+#define UTIL_PRINT_BITS(val, len, ...) do { \
+    const char* _sep = "|"; \
```
cnlohr: plx try to keep tabs instead of spaces.

But also...

please don't make macro functions this large.  when you do code size balloons. If you can pass in a const char * and a function that handles all this, that would be really good in comparison.

You can also use variadic functions if you want to maintain the usage of the  ...

cnlohr:
I'm ok with this register debugging configuration.

PR #748, opened 2025-11-08
Add missing ETH PHY reg definitions

kuwoyuki:


cnlohr:
@BogdanTheGeek can you review these changes? I am unfamiliar with this peripheral.

ch32fun/ch32v20xhw.h
```diff
@@ -5330,15 +5352,48 @@ then stop sending, 10=send pause frame periodically, 01=send pause frame once, t
 #define PHY_BMCR                0x00                                            /* Control Register */
```
BogdanTheGeek, line 5354: i think this section needs a bit of a description of how you are actually expected to use these offsets with: MIERGADR, MIRD, and MIWR. Other than that, it looks like a faithful implementation of the RM.

kuwoyuki: Added PHY reg access example

PR #754, opened 2025-11-11
Set _i2c_slave_struct.size1 and .size2 to volatile

squidgeefish:
This discourages the compiler from optimizing out the initialization of these fields. If this initialization is skipped, they default to 0, which causes `I2C1_EV_IRQHandler` to cycle endlessly since the data register will never be read due to the size check failing...

(This "optimization" may occur when the `i2c_registers` array is referenced outside of the `onWrite()` callback, such as in `main()`.)


cnlohr:
Thank you for finding and solving this issue.  CI passed. I merge.

PR #755, opened 2025-11-11
ch5xx: abolish FUNCONF_INIT_ANALOG

biemster:
`FUNCONFIG_INIT_ANALOG` was introduced with the initial port of the first ch5xx, the ch592, because I did not have a demo for the ADC yet and the ch32 `funAnalogInit()` uses `RCC` which ch5xx does not have.
We now have great ch5xx ADC demos thanks to @unicab369 (#727) so this macro can be removed.
(I also fixed the compiler warnings in the `adc_basic` demo)

### DIFF
--- a/examples_ch5xx/adc_basic/adc_basic.c
+++ b/examples_ch5xx/adc_basic/adc_basic.c
@@ -14,13 +14,13 @@ int main() {
 		adc_set_channel(ADC_TEMPERATURE_CHANNEL);
 		adc_set_config(ADC_FREQ_DIV_10, ADC_PGA_GAIN_2, 1);
 		adc_raw = adc_get_singleReading();
-		printf("\nTemperature: %ld mC\r\n", adc_to_mCelsius(adc_raw));
+		printf("\nTemperature: %d mC\r\n", (s16)adc_to_mCelsius(adc_raw));
 
 		//# get battery reading
 		adc_set_channel(ADC_VBAT_CHANNEL);
 		adc_set_config(ADC_FREQ_DIV_10, ADC_PGA_GAIN_1_4, 0);
 		adc_raw = adc_get_singleReading();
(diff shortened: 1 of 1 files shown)


cnlohr:
do you mind recasting your ints to avoid the warnings.  %ld is excessive.

biemster:
> do you mind recasting your ints to avoid the warnings. %ld is excessive.

done, casted them to `s16`.

cnlohr:
@biemster it should take less space if you just cast to an int.  Do you mind testing the code size and comparing between (s32) and (s16)?

Charles

biemster:
> @biemster it should take less space if you just cast to an int. Do you mind testing the code size and comparing between (s32) and (s16)?
> 
> Charles

I did not know that! I'll use this from now on. Further, this is not my demo, so I did not want to change it too much. I just did the minimal change to make the compiler warning go away.
When using `s16` it's 2136 bytes:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:        2136 B       448 KB      0.47%
             RAM:           0 B        32 KB      0.00%
```
using `s32` and `%ld` it's smaller (!), 2132 bytes:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:        2132 B       448 KB      0.46%
             RAM:           0 B        32 KB      0.00%
```
and `int` and `%d` it's also 2132 bytes:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:        2132 B       448 KB      0.46%
             RAM:           0 B        32 KB      0.00%
```

So actually my `s16` solution made it bigger :smiley: . Want to revert? (I really don't understand why I get these numbers though, my best guess is the compiler optimisation)

cnlohr:
I expected that s32 would make it smaller. I wouldn't revert, but if you're up for a new PR!

(in general using native types are faster than smaller types, because if you use smaller types, sometimes C needs to mask off upper bits)

PR #758, opened 2025-11-12
Update i2c_slave.c example to check i2c_registers in main()

squidgeefish:
Per @cnlohr's request in Discord, I've updated the example program to show that it's now safe to read from the `i2c_registers` array in `main()` rather than only in the `onWrite` callback.


cnlohr:
thanks!

PR #760, opened 2025-11-14
Implement constructor functions

doragasu:
Implement constructor functions and add an example.


cnlohr:
This is pretty good. Thanks!

PR #764, opened 2025-11-17
Clean up minichlink makefile

meganukebmp:
This change cleans up the Makefile a bit.

From @unicab369 on discord we know that minichlink compiles fine on Clang, so it's safe to not hardcode the compiler and instead pull it from $(CC)

Remove the platform (windows) CFLAGS and LDFLAGS and just override them in the conditional like it's done for macos.

Remove the .exe build instructions. Make on Windows will actually handle this for us. If we're building an executable it will append it with .exe. Likewise for executions, deletions and so on.

Add minichlink.dll to the default build to mirror the steps for *nix. Prior it had to be invoked explicitly.

Remove the makefile itself from rule dependencies. Doesn't hurt but makes no sense. How do we invoke a build step from the makefile without the makefile existing.

### DIFF
--- a/minichlink/Makefile
+++ b/minichlink/Makefile
@@ -58,4 +58,4 @@ inspect_bootloader : minichlink
 	riscv64-unknown-elf-objdump -S -D test.bin -b binary -m riscv:rv32 | less
 
 clean :
-	rm -rf $(TOOLS)
+	rm -rf $(TOOLS) minichlink.exe


minichlink/Makefile
```diff
@@ -39,18 +39,14 @@ endif
 
 all : $(TOOLS)
 
-# will need mingw-w64-x86-64-dev gcc-mingw-w64-x86-64
-minichlink.exe : $(C_S) $(H_S) Makefile
```
cnlohr, line 43: How does the .exe get generated in the new version?

minichlink/Makefile
```diff
@@ -8,10 +8,10 @@ H_S:=cmdserver.h funconfig.h hidapi.h libusb.h microgdbstub.h minichlink.h seria
 # gdb-multilib {file}
 # target remote :2345
 
-CFLAGS_WINDOWS:=-Os -s -Wall -D_WIN32_WINNT=0x0600 -I. -Wall -DCH32V003 -DMINICHLINK
-LDFLAGS_WINDOWS:=-L. -lpthread -lusb-1.0 -lsetupapi -lws2_32
 ifeq ($(OS),Windows_NT)
-	TOOLS:=minichlink.exe
+	TOOLS:=minichlink minichlink.dll
```
cnlohr, line 12: Doesn't this need to be minichlink.exe?

minichlink/Makefile
meganukebmp: Make on windows will actually handle this for us. Not specifying an extension leads to generating an exe.

meganukebmp: Likewise commands like `rm minichlink` will delete minichlink.exe. I can only assume this is intentionally done by mingw devs

minichlink/Makefile
meganukebmp: See above comment

cnlohr:
If I can get a signoff from @monte-monte then I think we should be fine.  (running CI)

minichlink/Makefile
monte-monte: make minichlink.exe is not only for windows. I use it to build windows executables on linux.

monte-monte:
How are you supposed to build minichlink.exe on linux now? What issue does this cleanup solves?

meganukebmp:
> How are you supposed to build minichlink.exe on linux now? What issue does this cleanup solves?

By specifying the OS and compiler in the env.

```
[nexrem@wyvern minichlink]$ make CC=x86_64-w64-mingw32-gcc OS=Windows_NT
x86_64-w64-mingw32-gcc -o minichlink minichlink.c pgm-wch-linke.c pgm-wch-isp.c pgm-esp32s2-ch32xx.c nhc-link042.c ardulink.c serial_dev.c pgm-b003fun.c minichgdb.c chips.c ch5xx.c -L. -lpthread -lusb-1.0 -lsetupapi -lws2_32 -O0 -g3 -Wall -Wno-unused-function -DCH32V003 -I. -DMINICHLINK 
x86_64-w64-mingw32-gcc -o minichlink.dll minichlink.c pgm-wch-linke.c pgm-wch-isp.c pgm-esp32s2-ch32xx.c nhc-link042.c ardulink.c serial_dev.c pgm-b003fun.c minichgdb.c chips.c ch5xx.c -L. -lpthread -lusb-1.0 -lsetupapi -lws2_32 -O0 -g3 -Wall -Wno-unused-function -DCH32V003 -I. -DMINICHLINK  -shared -DMINICHLINK_AS_LIBRARY
[nexrem@wyvern minichlink]$ ls
99-minichlink.rules  ch5xx.h  cmdserver.h  flake.nix    hidapi.h        Makefile        minichlink.c    minichlink.h   pgm-esp32s2-ch32xx.c  README.md     terminalhelp.h
ardulink.c           chips.c  default.nix  funconfig.h  libusb-1.0.dll  microgdbstub.h  minichlink.dll  nhc-link042.c  pgm-wch-isp.c         serial_dev.c  winbuild.bat
ch5xx.c              chips.h  flake.lock   hidapi.c     libusb.h        minichgdb.c     minichlink.exe  pgm-b003fun.c  pgm-wch-linke.c       serial_dev.h
```

This cleanup removes an unnecessary duplicate build rule and removes the need for a hardcoded compiler. As a result running make on the target system will pick the appropriate compiler based on what CC is set to (be it clang or gcc or something else)

One regression I just noticed is that when cross compiling the clean step does not remove the exe. I'll fix that

meganukebmp:
I'm unsatisfied with hardcoding the name anywhere but anything more complex will needlessly convolute what's already quite simple and readable. My only point of contention (and the one this PR resolves) is that there was a duplicate rule where one was unnecessary.

monte-monte:
I think you can make unhardcoded rule when OS is windows and leave separate ``minichlink.exe`` hardcoded rule for cross-compilation.

meganukebmp:
> I think you can make unhardcoded rule when OS is windows and leave separate `minichlink.exe` hardcoded rule for cross-compilation.

This is basically what it was before my changes, but it was redundant. We had the rules for minichlink and minichlink.exe being identical. Just running mingw gcc will automatically produce a binary suffixed with .exe.

monte-monte:
Please leave redundant ``minichlink.exe`` target. I am using it. It's not a big deal to have redundant stuff in Makefile - it's free.

cnlohr:
I am deferring to @monte-monte so I am closing this issue now.

meganukebmp:
Would a rework of this be of interest?

monte-monte:
@meganukebmp that's what I asked :) Improve windows side, but leave a separate rule for cross-platform build as is.

PR #766, opened 2025-11-19
Ch32x30x rtc test

unicab369:
added RTC example for Ch32x30x.
added a RTC helper that uses the same code to calculate the date time for CH5xx and CH32x30x.
update register_debug_utilities to change the bits order to MSB.


cnlohr:
In the future for future updates, please make the library functions static, but, for now, this seems fine.

PR #767, opened 2025-11-19
Ch32v30x iwdg example.

unicab369:


examples_v30x/iwdg_test/iwdg_test.c
```diff
@@ -0,0 +1,55 @@
...
+	printf("\n~ IWDG Test ~\n");
+
+	// set PA3 to input pullup
+	funPinMode(PA3, GPIO_CFGLR_IN_PUPD);
+
+	// reload value for 5 second timeout = (5 * 40KHz / 64) - 1 = 3124
+	IWDG_init(IWDG_Prescaler_64, 3124);
+
+	while(1) {
+		u8 read = funDigitalRead( PA3 );
```
cnlohr: please use `int` here or something like it.  Overspecifying type sizes produces more code. For instance, in many places on 32-bit systems if you specify `u8` instead of `u32` it has to produce more code because it has to mask off the bigger number

unicab369: hmmm, ok I didnt know that. thanks.

PR #768, opened 2025-11-19
Improve Non-Aligned Read Performance

PinkiePie1:
This update speeds up non-aligned reads. Commands such as ```minichlink -r + 0x03 1200```, which previously took several seconds to complete, now finish almost instantly. The improvement is modest on native Linux systems but much more noticeable when running under WSL.


BogdanTheGeek:
Looks like a good fix!

cnlohr:
@monte-monte can you review this?  I like these changes.

monte-monte:
@cnlohr seems good

cnlohr:
@PinkiePie1 can you fix the conversion to spaces you did?

cnlohr:
thanks

PR #772, opened 2025-11-20
Add 433 MHz transmitter example

sodoku:
I was playing around with 433Mhz and thought this might help someone. Feel free to reject if this is too basic.


unicab369:
can you add example how to receive the data?

sodoku:
> can you add example how to receive the data?

This example works with a simple transmit only board. I haven't used a receiver yet, but I can see if I can also make it work. Would you split this in two examples or put both in one?

unicab369:
if the code is too big, I would split it into 2 if not one is ok i think.

cnlohr:
Is it possible to do both send and receive from one part?  If so, that would be best.

And I am glad you kept things simple, but I do think the minimum viable product would be both send and receive.

cnlohr:
So definitely not a no, @sodoku but, maybe a "when you get TX/RX going.

Other than that the PR is good.

sodoku:
@cnlohr I managed to solder my receiver antenna and play around a bit with it. Reading how rc-switch does the receive and looking through the ch32fun examples, it looks like I will need to use interrupts for the input data change detection and for being able to measure time difference (`micros()` in Arduino). Imho that would make this example more complex.
What do you think?

cnlohr:
@sodoku on ch32fun, you use `SysTick->CNT`, and for comparing times you do something like:

```c
uint32_t start_time = SysTick->CNT;
while( 1 )
{
   // Do stuff
   if( TimeElapsed32( SysTick->CNT, start_time ) >= Ticks_from_Ms(150) )
   {
      // 150ms elapsed.
      start_time += Ticks_from_Ms(150);
   }
}
```

sodoku:
I spent hours on this and learned a lot, but somehow the timing values didn't make sense. I am not sure what I am doing wrong. I boiled my problem  down to a very simple example:

```c
int miliseconds = 15000;

uint32_t start_time = SysTick->CNT;

while ( 1 )
{
  // Do stuff
  uint32_t diff = TimeElapsed32( SysTick->CNT, start_time );
  if ( diff >= Ticks_from_Us( miliseconds ) )
  {
    start_time += Ticks_from_Us( miliseconds );
    printf( "%lu \n", diff / DELAY_US_TIME );
  }
}
```

This works an outputs:

```
+15000
+15000
+15000
+15000
+15000
+15000
+15000
+15000
```

However if I change `miliseconds` to a smaller value of `1500` since I need to count the timings in small values of microseconds (e.g 650), I get:

```
+1500
+1575
+1722
+1811
+1974
+2188
+2366
+2428
+2456
+2492
+2498
+2523
+2552
+2668
+2819
+3032
+3198
+3351
+3429
+3453
+3478
+3512
+3683
+3723
+3801
```

What am I doing wrong?

cnlohr:
You need to do the math on the uint32_t

but you need to compare based on int.

I.e.

```c
if( (int32_t)TimeElapsed( SysTick->CNT, endTime ) > 0 )
{
   endTime = SysTick->CNT + Ticks_from_Us( miliseconds );
}
```

or

```c
if( (int32_t)TimeElapsed( SysTick->CNT, endTime ) > 0 )
{
   endTime += Ticks_from_Us( miliseconds );
}
```

if you don't want your timebase to slip

sodoku:
I have given up with the 433 recieve (at least for now). I got none of the suggestions to work properly. The timings were never right. As a last resort, I wanted to make sure the receiver is not faulty. I tested it on an ESP32 with Arduino & rc-switch and it works fine on. I then tried to run the same code on wch arduino & rc-switch. Just had to make on simple change from

```c
attachInterrupt(this->nReceiverInterrupt, handleInterrupt, CHANGE);
```

to

```c
attachInterrupt(this->nReceiverInterrupt, GPIO_Mode_IN_FLOATING, handleInterrupt, EXTI_Mode_Interrupt, EXTI_Trigger_Rising_Falling); 
```

It compiles and runs, the interrupt fires, but rc-switch can't decode the timings.

I might look at receive again after I had some more time to learn about the CH32V003, ch32fun and lower level microcontroller programming.

cnlohr:
@sodoku We've changed a bunch around, I think this project would make a ton of sense to go into the `projects` folder.  If you're up for moving it there, I'd be happy to accept this!

sodoku:
@cnlohr I moved it to projects.

PR #773, opened 2025-11-20
Ch32v30x BKP test.

unicab369:


examples_v30x/bkp_test/ch32v30x_bkp.h
cnlohr, line 1: Unlike the other project, this makes sense I think to be in a header.

In the future please make these "static"

cnlohr:
In the future would be good if the header made its functions `static` but this is fine.

PR #774, opened 2025-11-20
Ch32v30x crc test.

unicab369:
pretty straight forward.

### DIFF
--- a/examples_v30x/crc_test/crc_test.c
+++ b/examples_v30x/crc_test/crc_test.c
@@ -2,41 +2,7 @@
 
 #include "ch32fun.h"
 #include <stdio.h>
-
-static void CRC_init(void) {
-	// Enable CRC clock
-	RCC->AHBPCENR |= RCC_AHBPeriph_CRC;
-}
-
-// Calculate CRC for a single 32-bit value
-static u32 CRC_calculate32(u32 data) {
-	// Reset CRC unit
-	CRC->CTLR = CRC_CTLR_RESET;
-	
-	// Write data to trigger CRC calculation
-	CRC->DATAR = data;
-	
-	// Wait for CRC calculation
-	u32 timeout = 10000;
-	while (CRC->DATAR == 0 && --timeout);
-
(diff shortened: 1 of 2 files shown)


examples_v30x/crc_test/ch32v30x_crc.h
```diff
@@ -0,0 +1,34 @@
...
+
+// Calculate CRC for a single 32-bit value
+u32 CRC_calculate32(u32 data) {
+	// Reset CRC unit
+	CRC->CTLR = CRC_CTLR_RESET;
+	
+	// Write data to trigger CRC calculation
+	CRC->DATAR = data;
+	
+	// Wait for CRC calculation
```
cnlohr: This is impractical.  Please add the while(...) to wait for the completion.  And verify that it works.

examples_v30x/crc_test/ch32v30x_crc.h
```diff
@@ -0,0 +1,34 @@
+#include "ch32fun.h"
```
cnlohr: Please just put these functions in the main file.

cnlohr:
Just have two comments.  Once addressed, looks good.

examples_v30x/crc_test/ch32v30x_crc.h
unicab369: should I put this in a lib bc I also want to use this with v20x and v30x?

examples_v30x/crc_test/ch32v30x_crc.h
unicab369: unfortunately, the CRC register doesn't not have a flag to check when the calculation is completed. I added a timeout.

examples_v30x/crc_test/ch32v30x_crc.h
cnlohr: Oh. I didn't know.  In that case, yes, you can put them in a lib.

cnlohr:
Looks good!

PR #775, opened 2025-11-20
Ch32v30x Random Number Genearator test.

unicab369:
pretty straight forward as well.

### DIFF
--- a/examples_v30x/rng_rand_test/ch32v30x_rng.h
+++ b/examples_v30x/rng_rand_test/ch32v30x_rng.h
@@ -1,6 +1,6 @@
 #include "ch32fun.h"
 
-void RNG_init(void) {
+static void RNG_init(void) {
 	// Enable RNG clock
 	RCC->AHBPCENR |= RCC_AHBPeriph_RNG;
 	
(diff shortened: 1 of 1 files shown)


examples_v30x/rng_rand_test/ch32v30x_rng.h
```diff
@@ -0,0 +1,26 @@
...
+// Get a random number (blocking)
+u32 RNG_rand(void) {
+	// Wait until data is ready
+	while(!(RNG->SR & RNG_SR_DRDY)) {
+		// Handle seed error
+		if(!(RNG->SR & RNG_SR_SECS)) continue;
+
+		RNG->CR &= ~RNG_CR_RNGEN;		// Disable RNG
+		RNG->SR = 0;					// Clear error flags
+		Delay_Us(1);					// Wait
```
cnlohr: Cannot wait.  Find out what takes time, and what you have to do to make this as fast and clean as possible.  Other people will be downstream.  this needs to be high quality.

Also since we are doing another round, can these functions be `static`?

cnlohr:
See above comment.

examples_v30x/rng_rand_test/ch32v30x_rng.h
unicab369: I'm not sure what causes the wait, if I read the register without waiting it doesn't return anything.

unicab369: Maybe the application code has to keep refresh the reading until they get a number

unicab369: actually, nvm. I added the delay bc of the while loop, the while loop should block until it finds a new number.

cnlohr:
Looks good

PR #776, opened 2025-11-20
HX711 Load Cell Amplifier Example

AlexanderMandera:
This example provides a library for the HX711 load cell amplifier board, mainly used for weighing scales in different forms.
It does not use any floating point math to improve performance.

Any suggestions for improvements or optimizations are appreciated.
I needed to put the hx711_shift_in method into RAM to work properly.
The whole demo is 4628 bytes, but maybe something can be cut down?


cnlohr:
1. The name of the project should include "load cell amplifier" or something like that in its name.  I.e. instead of "hx711_demo" it should be "load_cell_amplifier_lx711" or something like that.
2. There needs to be more explanation as to what is going on and how it is being used in the code.
3. Can you include a schematic in the readme.

AlexanderMandera:
Thanks for the feedback.
I can have a look at the weekend and make the adjustments.

cnlohr:
lmk when you get further.

cnlohr:
We are now trying out a new mechanism for more full projects for ch32fun. This seems like something that would be good to put in the `projects` folder https://github.com/cnlohr/ch32fun/tree/master/projects

cnlohr:
@AlexanderMandera up for putting this into the `projects/` folder? I'd love to accept it there.

AlexanderMandera:
Hi @cnlohr 
Sorry for the delay, I was quite busy the last few weeks, I will move the project in the next few days and let you know :)

cnlohr:
@AlexanderMandera any progress?

AlexanderMandera:
Hi @cnlohr
thanks for reminding me, I just was very busy with work and didn't have time to do the changes until now.
I hope it is fine now :)

cnlohr:
this looks good.

PR #777, opened 2025-11-20
added Ch32v30x PVD test.

unicab369:
added lib_pvd that share code between the v00x to v30x.

### DIFF
--- a/extralibs/lib_pvd.h
+++ b/extralibs/lib_pvd.h
@@ -2,7 +2,7 @@
 
 #include "ch32fun.h"
 
-void PVD_init(u8 threshold) {
+static void PVD_init(u8 threshold) {
 	if (threshold > PVD_MAX_THRESHOLD_LVL) threshold = PVD_MAX_THRESHOLD_LVL;
 
 	// Enable PWR clock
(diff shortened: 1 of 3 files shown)


extralibs/lib_pvd.h
```diff
@@ -0,0 +1,26 @@
+// Tested with CH32V002, CH32V006, and CH32V303
+
+#include "ch32fun.h"
+
+void PVD_init(u8 threshold) {
```
cnlohr: These should be static.

extralibs/lib_pvd.h
```diff
@@ -0,0 +1,26 @@
...
+
+	// Enable PVD
+	PWR->CTLR |= PWR_CTLR_PVDE;
+	
+	// Clear the existing PLS bits and set new threshold
+	PWR->CTLR = (PWR->CTLR & ~PWR_CTLR_PLS) | (threshold << 5);
+}
+
+// Get threshold setting: the PLS[1:0] bits
+u8 PVD_getThreshold() {
```
cnlohr: in general, make these `int`s so that the compiler doesn't do any extra masking.

Here, it's fine it's not a big deal, and the compiler will produce fine code, but in other cases it may not.

cnlohr:
@unicab369 I really like this, if you can address both comments, that would be great and then I can merge it.

cnlohr:
Looks good

PR #783, opened 2025-11-23
Introduce CH32H41x

VynDragon:
adds ch32h417, 415, 416


minichlink/minichgdb.c
```diff
@@ -322,6 +322,8 @@ int RVReadMem( void * dev, uint32_t memaddy, uint8_t * payload, int len )
 		fprintf( stderr, "Error: Can't alter halt mode with this programmer.\n" );
 		exit( -6 );
 	}
+	if (memaddy < 0x08000000)
```
cnlohr: We can't just do this.  There's situations where you want to read/write to the actual 0x00 base address, like if you are playing with bootloaders.

VynDragon: I forgot to remove that (was never intended to be final), however it's necessary to be done in some way to debug that ch32 so not sure how to deal with that

VynDragon: removed

cnlohr:
Looks good!

Issue #625, opened 2025-06-22
Missing defines


paulwedeck:
CAN_TypeDef misses the TTCTLR and TTCNT registers.

cnlohr:
@paulwedeck  Which processor?

paulwedeck:
ch32v20x and ch32v30x
The other ones don't have CAN controllers afaik.

cnlohr:
I have no way of testing CAN, but if you're up for testing it I would really appreciate a PR.  If you need help making a PR please let me know.

PR #787, opened 2025-11-25
Add missing fields to CAN struct. Add CAN for ch32v103

monte-monte:
Fixes https://github.com/cnlohr/ch32fun/issues/625, also adds missing CAN definitions to ch32v103.


cnlohr:
Looks good.

Issue #628, opened 2025-06-25
[minichlink] CH32V003 Firmware read failed


adimiti:
When attempting to read the entire flash memory, the operation fails:

```
minichlink -r dump.bin flash 0x4000
Found WCH Link
WCH Programmer is LinkE version 2.12
Full Chip Type Reply: [8] 82-0d-05-09-00-31-05-00-03
Detected: CH32V003
Setup success
Flash Storage: 16 kB
Part UUID    : 60-b0-ab-cd-c8-b8-bc-3b
PFlags       : ff-ff-ff-ff
Part Type (B): 07-13-bb-91
Read protection: disabled
Interface Setup
Fault on op (DMABSTRACTS = 08000302) (100) (Execption executing Abstract Command) DMSTATUS: 004c0382
Fault on DefaultReadBinaryBlob
Fault reading device
```

But partial (e.g., 0x0–0x3FFF) succeeds.

```
minichlink -r dump.bin flash 0x3FFF
Found WCH Link
WCH Programmer is LinkE version 2.12
Full Chip Type Reply: [8] 82-0d-05-09-00-31-05-00-03
Detected: CH32V003
Setup success
Flash Storage: 16 kB
Part UUID    : 60-b0-ab-cd-c8-b8-bc-3b
PFlags       : ff-ff-ff-ff
Part Type (B): 07-13-bb-91
Read protection: disabled
Interface Setup
Read 16383 bytes
```

TommyMurphyTM1234:
Are you able to read the firmware back using the WCH flash tool or OpenOCD perhaps - just for comparison?

adimiti:
Yes, same chip with same LinkE , from another laptop with windows 11:

![Image](https://github.com/user-attachments/assets/75ed4d34-0836-4496-821d-faf2c5ccae9d)

PinkiePie1:
I believe the full flash range is from 0x0000 to 0x3FFF, and 0x4000 is reserved according to the datasheet. The WCH flash tool also shows nothing at 0x4000, so that might be why minichlink refuses to access it.

cnlohr:
I've also seen this same behavior.  I think it's not a bug?  What happens if you only read the bottom 0x3ff0?

adimiti:
> I believe the full flash range is from 0x0000 to 0x3FFF, and 0x4000 is reserved according to the datasheet. The WCH flash tool also shows nothing at 0x4000, so that might be why minichlink refuses to access it.

[...]

PR #788, opened 2025-11-26
Fix reads on memory boundary

monte-monte:
Fixes https://github.com/cnlohr/ch32fun/issues/628
Issue was with autoincrement autoincrementing past available memory, and brogbuf failing to execute.

### DIFF
--- a/minichlink/chips.c
+++ b/minichlink/chips.c
@@ -806,3 +806,8 @@ const struct RiscVChip_s* FindChipISP(uint16_t chip_id)
 	}
 	return chip;
 }
+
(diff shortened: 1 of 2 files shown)


cnlohr:
I have no opinions on this. Any reason not to merge?

monte-monte:
@cnlohr this resolves old bug. The only other way is to disable autoincrement completely or ignore errors in ``waitfordoneop`` in ``defaultreadblob`` function.
To see difference on practice the easiest and the quickest way is to run ``./minichlink -r + bootloader 1920`` with v003 connected before and after this commit.

PR #790, opened 2025-11-26
Enable build of ch32fun and minichlink on FreeBSD

adventureloop:


ch32fun/ch32fun.mk
```diff
@@ -11,6 +11,8 @@ PREFIX_DEFAULT:=riscv64-elf
 
 ifneq ($(shell $(WHICH) riscv64-unknown-elf-gcc 2>$(NULLDEV)),)
 	PREFIX_DEFAULT:=riscv64-unknown-elf
+else ifneq ($(shell $(WHICH) riscv32-unknown-elf-gcc 2>$(NULLDEV)),)
+	PREFIX_DEFAULT:=riscv32-unknown-elf
```
cnlohr, line 15: Glad to have another fallback

ch32fun/ch32fun.mk
```diff
@@ -394,7 +376,7 @@ gdbclient :
 	gdb-multiarch $(TARGET).elf -ex "target remote :3333"
 
 clangd :
-	make clean
+	$(MAKE) clean
```
cnlohr, line 379: Can you put a MAKE?=make earlier? so on old Windows Make versions it still works?

minichlink/Makefile
```diff
@@ -44,10 +49,10 @@ minichlink.exe : $(C_S) $(H_S) Makefile
 	x86_64-w64-mingw32-gcc -o $@ $(C_S) $(LDFLAGS_WINDOWS) $(CFLAGS_WINDOWS)
 
 minichlink : $(C_S) $(H_S) Makefile
-	gcc -o $@ $(C_S) $(LDFLAGS) $(CFLAGS) $(INCS)
+	cc -o $@ $(C_S) $(LDFLAGS) $(CFLAGS) $(INCS)
```
cnlohr, line 52: This may not work on some Windows crossdev systems.

minichlink/minichlink.c
```diff
@@ -67,18 +71,22 @@ void * MiniCHLinkInitAsDLL( struct MiniChlinkFunctions ** MCFO, const init_hints
 		{
 			fprintf( stderr, "Found WCH Link\n" );
 		}
+#ifndef __FreeBSD__
```
cnlohr, line 74: Please try to add support for these.   Disabling functionality on specific platforms should only be done as a later "I give up"

ch32fun/ch32fun.c
```diff
@@ -1755,6 +1495,7 @@ WEAK int _write(int fd, const char *buf, int size)
 	return size;
 }
 
+#undef putchar
```
cnlohr, line 1498: #ifdef putchar
#undef putchar
#endif

minichlink/Makefile
adventureloop: my bad, I will add platform detection

minichlink/minichlink.c
adventureloop: esp32 and b0003 need "hidapi_thread_pthread.h" which isn't provided by the port (package basically) right now. The header doesn't seem to be in the public hidapi interface, I'm not sure if other platforms package this with their hidapi package, I need to get a debian machine to check.

adventureloop:
Ah ha, I got it by updating hidapi in minichlink and pulling in the missing header.

I'll create a different review for that, do you know if a hidapi update is safe?

minichlink/minichlink.c
cnlohr: I am not sure what you are referring to.  hidapi.c/.h is provided alongside minichlink.  They contain all of hidapi on all supported platforms.

cnlohr:
> Ah ha, I got it by updating hidapi in minichlink and pulling in the missing header.
> 
> I'll create a different review for that, do you know if a hidapi update is safe?

There should be no additional header. the provided hidapi.h/.c are fully merged for all platforms.

cnlohr:
I think? this has been resolved by #872 and #873

tsutsui:
> I think? this has been resolved by #872 and #873

Unfortunately, I'm afraid NetBSD and FreeBSD have completely different APIs for USB hid devices.

However it looks there is a pull request to add FreeBSD's `hidraw` support to upstream hidapi,
https://github.com/libusb/hidapi/pull/730
so it would be better to use a newer hidapi library (from FreeBSD ports etc.) with the new FreeBSD-native APIs so that no need to tweak libusb portion (like `hidapi_thread_pthread.h`) in locally pulled ch32fun/minichlink/hidapi.c.

adventureloop:
Cool thanks for that pointer, I will follow that upstream

cnlohr:
If someone wants to update our hidapi, I'm happy to let that one through!

PR #791, opened 2025-11-27
Ch5xx spi st7735 demo

unicab369:


examples_ch5xx/spi_st7735_test/funconfig.h
```diff
@@ -0,0 +1,15 @@
...
+#define _FUNCONFIG_H
+
+#define FUNCONF_USE_HSI           0 // CH5xx does not have HSI
+#define FUNCONF_USE_HSE           1
+#define CLK_SOURCE_CH5XX          CLK_SOURCE_PLL_60MHz // default so not really needed
+#define FUNCONF_SYSTEM_CORE_CLOCK 60 * 1000 * 1000     // keep in line with CLK_SOURCE_CH5XX
+
+#define FUNCONF_DEBUG_HARDFAULT   0
+#define FUNCONF_USE_CLK_SEC       0
+#define FUNCONF_INIT_ANALOG       0 // ADC is not implemented yet
```
biemster: this doesn't exist anymore, please leave it out from now

cnlohr:
@unicab369 please put the header files here into the folder of the demo.  They're not ready for extralibs.

cnlohr:
Also, for all of them, even if not in common, please start using `int` or `uint32_t` etc... instead of `u8` when `int` would do.  It really makes the code more bloated and slower unnecessarily to use `u8` or `u16` unless you are actually doing something with arrays, etc. that are of the specific sizes.

unicab369:
@cnlohr I moved the files and make sure it built. I have to revisit this later to update the variable types and make sure it's still working.

cnlohr:
Sorry, I also realized a few things.

1. I believe we already added a struct definition for the SPI for the CH5xx
2. Please add documentation for where the font5x8 came from.

PR #794, opened 2025-11-28
Ch5xx I2C Cleanup.

unicab369:


cnlohr:
1. In the future if you can write summaries as to why you want to do cleanups. Or any time there's more to something if you can explain it.
2. If you are going to be doing this, I also recommend switching from. u16 to things like int so the compiler doesn't have to emit extra code to mask off bits.

cnlohr:
This is conflicting now, maybe you can rebase this?

unicab369:
idk what's going on with this branch. I think I got it mixed up on my local.

unicab369:
@cnlohr I think it's updated.

PR #795, opened 2025-11-29
fix: fsusb.c warnings

kuwoyuki:


extralibs/fsusb.c
```diff
@@ -417,6 +417,8 @@ void USBFS_IRQHandler()
 						if( len > USBFS_SetupReqLen )
 							len = USBFS_SetupReqLen;
 						ctx->USBFS_SetupReqLen = len;
+
+						break;
```
cnlohr, line 421: I'm 99% sure this will break a lot of things, this relies on the code falling through from above. You will need to find some other way with fallthrough to inform GCC, like `__attribute__ ((fallthrough));`

cnlohr:
See my comment in code.

cnlohr:
Actually, it might be a bug? can you test/confirm both ways?

kuwoyuki:
> Actually, it might be a bug? can you test/confirm both ways?

Both work. I don't believe the address should be set immediately after preparing a descriptor?  
`ctx->USBFS_DevAddr` might have garbage data until setup packet is received and gets overwritten.

cnlohr:
I concur.  It looks like a bug.  I will merge this.

Issue #694, opened 2025-09-06
ch5xx improvements


biemster:
While adding support for the ch5xx series I cut some corners here and there, and left some optimisations for later. This ticket will keep track of these things so they are not forgotten, but it does not mean _I_ have to implement them! So please feel free to open a PR if you want to work on, or have a solution for any of these.

- [ ] Startup: The startup of ch32 is taken, while the ch5xx series are slightly different. It works fine, but there might be room for optimisation
- [ ] Startup: Add proper highcode sections for all ch5xx, currently only ch592 has basic implementation
- [ ] Startup: The ch5xx startup has one extra 0 entry at the beginning of the vector compared to the ch32 startup. This causes the magic bytes to be on an index one off compared to the EVT
- [ ] SystemInit: Not all clock cfg settings are tested, and some are giving odd behavior (in the program itself or in the debug interface)
- [ ] SystemInit: The flash configs that go together with the clock speed are not optimized, or documented. It is possible that for lower clock speeds there are unnecessary wait states configured, slowing down execution.
- [ ] iSLER: AccessAddress should be configurable, it is now hardcoded to BED6
- [ ] iSLER: The chips supporting S2/S8 show weird behavior on S8 mostly
- [ ] iSLER: There are several timers based on frame length, and some are way off resulting in unnecessary idle times / power draw.
- [ ] iSLER: For any BLE stack we need the PDU of the received frame
- [ ] iSLER: RSSI on received frames seems to return 0x02 almost exclusively, there must be something off there
- [ ] GPIO: The `GPIO_SetBits and _ResetBits` should use `_CLR` and `_SET` and `=` instead of `|=`, to make it a good clip faster.

cnlohr:
These would all be wonderful.

biemster:
This is starting to wrap up nicely! The points that are left are of quite low priority to me, and many of them are not even issues. If this ticket is still open long after this comment without an update, it can be closed.

PR #798, opened 2025-12-01
iSLER fixes for #694

biemster:
There are several items in #694 that concern iSLER, and since I want to put some time in WiNot I'm going to tackle those as well in this PR. It'll take a while to do them all, so it's draft for now.


biemster:
turns out RSSI on 570 is in a different register, I have to find which. This can be merged since everything else works fine, but we could also wait until I found this nit.

cnlohr:
all looks good.

Thanks.

PR #803, opened 2025-12-07
ch5xx: revert ed1b257 for 571/3 and 582/3, they don't have GPIO _SET

biemster:
Today I found that my 582's don't know how to blink anymore, turns out they don't have the gpio _SET register so ed1b257 broke `GPIO_SetBits` on them.
I've tested 570/2 and 59x which do work with _SET (and I assume 585 does too), and also tested on 573 and 582 which don't have _SET.


cnlohr:
Where does this issue stand?

PR #804, opened 2025-12-08
HSUSB flow control for OUT (Rx) transfers

doragasu:
Implements flow control for data OUT transfers.

Tested on CH32V305, should also work on devices with `USBHS_IMPL==2`, but I have no one to test.

I would also like to implement hardware double buffering to improve speed, but until I get the time to do that, this works.

### DIFF
--- a/extralibs/hsusb.c
+++ b/extralibs/hsusb.c
@@ -1,6 +1,7 @@
 #include "hsusb.h"
 #include "ch32fun.h"
 #include <string.h>
+#include <stdio.h>
 
 struct _USBState USBHSCTX;
 volatile uint8_t usb_debug = 0;
@@ -25,6 +26,10 @@ void * fast_memcpy( void *dst, void *src, uint32_t size )
 // #define copyBuffer memcpy
 #else
 #define copyBuffer memcpy
+#endif
+
+#if (FUSB_OUT_FLOW_CONTROL > 0) && (FUSB_USER_HANDLERS < 1)
+#error FUSB_OUT_FLOW_CONTROL requires FUSB_USER_HANDLERS
 #endif
 
 #if (USBHS_IMPL==2)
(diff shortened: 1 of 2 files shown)


monte-monte:
Nice!

cnlohr:
I don't think that's the best way to go about this.

No changes to hsusb.c are needed.

Some of your cleanup is good, but not particularly needed.

Also for out, there are multiple meanings for when you would want to nak/pause/go.  I have an example here:  https://github.com/cnlohr/cnhardware/blob/master/xmastree/mainboard/firmware-bandc/firmware.c#L192-L205

doragasu:
While this might not be the best way to implement flow control, I cannot see how this can be done without modifying hsusb.c. As it is right now, it is automatically ACKing all transfers, causing the host to immediately send a new one (if available).

But I must be missing something, because I neither understand how the code you are linking is supposed to pause the host. If it is supposed to achieve it using the `USBHS_Endp_Busy` flag, that's the first thing I tried and doesn't work, the driver only uses it for IN transfers, completely ignores it for OUT transfers.

doragasu:
Oh I see, your example code does use the `hsusb_v30x` driver, that in fact, is NAKing all data OUT transfers by default, and using the `USBHS_Endp_Busy` flag: https://github.com/cnlohr/ch32fun/blob/d92331ce37c8a4d9bb3f90feb7c81be566623883/extralibs/hsusb_v30x.c#L507

The `hsusb` driver doesn't do that, it ACKs all OUT transfers and completely ignores the `USBHS_Endp_Busy` flag. If you prefer, I could refactor this to behave in a similar way to the `hsusb_v30x` driver.

cnlohr:
hm I didn't realize they had diverged?  

How much effort do you think it would be to try to more unify them?

Do you have a preference over which way you think would be better for other people down the line?

doragasu:
I think `hsusb` is already an effort to unify different USB stacks, so maybe `hsusb_v30x` can be just removed (and example projects using it adapted to `hsusb`)? I think @monte-monte will have a much better insight about this.

The current way of handling flow control for OUT transfers in `hsusb_v30x` seems OK for me, I could use a similar approach, I would just have to NAK all OUT data transfers as `hsusb_v30x` does, remove the conditional compilation stuff I added, remove the `USBHS_RxReady()` function (or maybe convert it to a macro for convenience), and I think that's all.

monte-monte:
hsusb was made by adapting fsusb for HS while checking how they do things in EVT. I like the proposed way in this PR. not sure why cnlohr is against it.

cnlohr:
Only issue I have is I am working on a project now that leverages `hsusb_v30x` but I should probably just move it along. We can accept this and I can switch it up when I get around to it.

What is the current v305 example that uses `hsusb`?

cnlohr:
(Note question even though this is merged)

doragasu:
`examples_usb/USBHS/usbhs_device`

PR #810, opened 2025-12-16
Added HW AES example

monte-monte:
Works on all ch5xx chips except ch570. Need more testing. Probably can move AES HW struct to ``ch5xxhw.h``? Also maybe someone can add sw AES for comparison.

### DIFF
--- a/examples_ch5xx/aes/README.md
+++ b/examples_ch5xx/aes/README.md
@@ -1,6 +1,6 @@
 # Hardware AES
 
-This examples shows how to use hardware AES encoder/decoder on ch5xx chips (the ones with BLE). All tested chips have this example working, except ch570. Enen though it has a radio and can be used for BLE in some capacity, it seems that AES module is disabled/missing/broken on this chip. This results in that the BLE pairing (bonding) fails.
+This example shows how to use hardware AES encoder/decoder on ch5xx chips (the ones with BLE). All tested chips have this example working, except ch570. Even though it has a radio and can be used for BLE in some capacity, it seems that AES module is disabled/missing/broken on this chip. This results in the BLE failing in the pairing (bonding) operation.
 
 ## Registers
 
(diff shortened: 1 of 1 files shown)


cnlohr:
Please add a note at the top of the main .c file explaining that this does not work with the ch570 before we accept.

monte-monte:
@cnlohr I didn't want to yesterday, but today wrote a brief readme.

examples_ch5xx/aes/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+``void AES_DevAESDec(uint32_t *key, uint32_t * plaintextData, uint32_t * decryptData)``
+
+``void AES_DevAESEnc(uint32_t *key, uint32_t * plaintextData, uint32_t * encryptData)``
+
+``void AES_DevPktDec(uint32_t param_1, uint32_t * param_2)``
+
+``void AES_DevPktEnc(uint32_t param_1, uint32_t * param_2)``
+
+The first two were REd into one ``doAES`` function in this example. Another two seem to be for bulk packet processing, maybe involving some DMA, but they haven't been reverse engineered yet. They use 4 registers for unknown purposes, and these registers are yet to be named.
```
TommyMurphyTM1234, line 17: I presume that "REd" means "reverse engineered"?
If so, then it's probably better to spell it out for clarity?

examples_ch5xx/aes/README.md
```diff
@@ -0,0 +1,23 @@
+# Hardware AES
+
+This examples shows how to use hardware AES encoder/decoder on ch5xx chips (the ones with BLE). All tested chips have this example working, except ch570. Enen though it has a radio and can be used for BLE in some capacity, it seems that AES module is disabled/missing/broken on this chip. This results in that the BLE pairing (bonding) fails.
```
TommyMurphyTM1234: Typos:

- "This examples" should be "This example"
- "Enen" should be "Even"
- "This results in that the BLE pairing (bonding) fails" should be "This results in the BLE pairing (bonding) failing"

examples_ch5xx/aes/README.md
monte-monte: really?..

cnlohr: RE'd is ok.  (But just for clarity you do usually use an apostrophe.)

The above would be good to fix.

PR #811, opened 2025-12-19
Working on esig, and CRC, and rx filter for ch570/ch572

cnlohr:


extralibs/iSLER.h
```diff
@@ -1,3 +1,44 @@
...
+		while(!rx_ready);
+		// Receive packet code.
+
+*/
+
+// 2025-12-25 function rename.
+#define RFCoreInit iSLERInit
+#define Frame_TX   iSLERTX
+#define Frame_RX   iSLERRX
+#define ReadRSSI   iSLERRSSI
```
biemster: maybe immediately update the demos as well in this PR?

PR #816, opened 2025-12-26
USBFS/HS bulk echo demo

biemster:
Add USB FS/HS echo examples on how to use Bulk endpoints and interact with them from the host using python scripts.

### DIFF
--- a/examples_usb/USBFS/usbfs_bulk_echo/testtop.bulk/Makefile
+++ b/examples_usb/USBFS/usbfs_bulk_echo/testtop.bulk/Makefile
@@ -1,0 +1,13 @@
+all : testtop
+
+# to make CI happy.
+testtop.bulk.elf : testtop.c
+	gcc -o $@ $^ -I../../../minichlink -ludev -lusb-1.0
+
+testtop : testtop.c
+	gcc -o $@ $^ -I../../../minichlink -ludev -lusb-1.0
+
+
+clean :
+	rm -rf *.o *~ testtop
+
--- a/examples_usb/USBFS/usbfs_bulk_echo/testtop.bulk/os_generic.h
+++ b/examples_usb/USBFS/usbfs_bulk_echo/testtop.bulk/os_generic.h
@@ -1,0 +1,521 @@
+#ifndef _OS_GENERIC_H
+#define _OS_GENERIC_H
+/*
+	"osgeneric" Generic, platform independent tool for threads and time.
+		Geared around Windows and Linux. Designed for operation on MSVC,
+		TCC, GCC and clang.  Others may work.
+
+    It offers the following operations:
+
+	Delay functions:
+		void OGSleep( int is );
+		void OGUSleep( int ius );
+
+	Getting current time (may be time from program start, boot, or epoc)
+		double OGGetAbsoluteTime();
+		double OGGetFileTime( const char * file );
+
+	Thread functions
+		og_thread_t OGCreateThread( void * (routine)( void * ), void * parameter );
+		void * OGJoinThread( og_thread_t ot );
+		void OGCancelThread( og_thread_t ot );
+
+	Mutex functions, used for protecting data structures.
+		 (recursive on platforms where available.)
+		og_mutex_t OGCreateMutex();
+		void OGLockMutex( og_mutex_t om );
+		void OGUnlockMutex( og_mutex_t om );
+		void OGDeleteMutex( og_mutex_t om );
+
+	Always a semaphore (not recursive)
+		og_sema_t OGCreateSema(); //Create a semaphore, comes locked initially.
+          NOTE: For platform compatibility, max count is 32767
+		void OGLockSema( og_sema_t os );
+		int OGGetSema( og_sema_t os );  //if <0 there was a failure.
+		void OGUnlockSema( og_sema_t os );
+		void OGDeleteSema( og_sema_t os );
+
+	TLS (Thread-Local Storage)
+		og_tls_t OGCreateTLS();
+		void OGDeleteTLS( og_tls_t tls );
+		void OGSetTLS( og_tls_t tls, void * data );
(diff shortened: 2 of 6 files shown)


biemster:
Both FS and HS demos work now, but the toggles in InRequest and DataOut are not consistent between FS/HS and look sus, they should not be necessary at all. While this works, I don't think it's ready to merge yet since we might want to understand first why these toggles are needed.

biemster:
ok now it's even echoing wrong data back, it worked just before I committed this :exploding_head: 
edit: that was only USBFS, and that is fixed by using copy in SendEndpoint.

biemster:
This is ready now, although then `extralibs/[fh]susb.c` changes could use a thumbs up from @monte-monte ?

examples_usb/USBFS/usbfs_bulk_echo/usbfs_bulk_echo.py
```diff
@@ -0,0 +1,70 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-b', '--bootloader', help='Reboot to bootloader', action='store_true')
+    args = parser.parse_args()
+
+    if device is None:
+        print("ch585 not found")
+        exit(0)
+
+    # Set the active configuration. With no arguments, the first configuration will be the active one
+    device.set_configuration()
```
monte-monte: Probably that's what was causing you issues? I think set configuration is done at enumeration and shouldn't be normally needed to open a device.

biemster:
While the DATAx reset on SET CONFIGURATION requests did fix the initial issues, and the DATAx bits should be reset on such a request, according to the spec the hardware should handle this. It is possible the bits are not reset because the same config is requested that is already active.

Removing `set_configuration` from the python scripts also fixes the issue, but I'm now not sure if the DATAx reset from `[fh]susb.c` is appropriate, or should be removed from this PR.

cnlohr:
@biemster  I really don't want to accept this without .c topside files as well, as I do not want the expectation to test any part of this to require python.

You can copy them from here:  https://github.com/cnlohr/ch32fun/tree/master/examples_v30x/usbdevice_hs/testtop.bulk  they should work with minimal changes.

biemster:
Ok I'll add the topside C.
I did not do that yet because I'm also still waiting for a decision on the manual toggle reset on SET CONFIGURATION, we should discuss that at some point.

cnlohr:
I'm not familiar with the SET CONFIGURATION discussion.  I can leave that up to you and @monte-monte to decide.  If they agree, then let's merrrrge.

biemster:
I've added the topside C, so qua completeness this is ready to merge. There are however 2 issues remaining, the SET CONFIGURATION issue:
https://github.com/cnlohr/ch32fun/pull/816/changes#diff-e2fd352121f2e5cc694f3714e64615a2f734f0946df5e215688c0de4aa4a80f6R449
and
https://github.com/cnlohr/ch32fun/pull/816/changes#diff-54349e1b91b342d7ce8a2b9f78e8307b5b34dd186af41ea14f956fb933da6234R265,

and the issue where GET STATUS from `lsusb` is timing out, due to `ctrl0buf` being a different pointer than `ctx->pCtrlPayloadPtr`. A simple `ctx->pCtrlPayloadPtr = ctrl0buff;` at the end of the `case USB_GET_STATUS` resolves that, but I'd like to hear from @monte-monte what should be done about both issues.

monte-monte:
About SET_CONFIGURATION:
- evt doesn't have this
- it shouldn't be a problem in normal use. but you've found an edge case and then found a solution.
- I couldn't think how it could break normal things, so maybe we can just add it for situations like this. but then...it seems like fixing a bug that is not on the hw side...but maybe I'm wrong.



> due to `ctrl0buf` being a different pointer than `ctx->pCtrlPayloadPtr`

why is it different, though?

biemster:
> About SET_CONFIGURATION:
> * I couldn't think how it could break normal things, so maybe we can just add it for situations like this. but then...it seems like fixing a bug that is not on the hw side...but maybe I'm wrong.

From what I've read in libusb issues there are devices that have this issue, and there are some that don't. Also it only happens when a configuration is set that was already active. The `chprog.py` flasher actually would have triggered this, if the bootloader also was susceptible to the issue that is remedied here. I do agree it's a bit of an edge case, but I also think it will prevent at least some users from encountering it. That said, I'd be fine with not including it too, since the hardware should do this. The usbfs library is correct as is.

> > due to `ctrl0buf` being a different pointer than `ctx->pCtrlPayloadPtr`
> 
> why is it different, though?

This I really don't understand, I was hoping you would :)

biemster:
`ctx->pCtrlPayloadPtr` is zero at that point I just found out, maybe that's the issue?

biemster:
Also, it's part of this switch:
```C
/* usb standard request processing */
switch( USBFS_SetupReqCode )
{
	/* get device/configuration/string/report/... descriptors */
	case USB_GET_DESCRIPTOR:
	{
		const struct descriptor_list_struct * e = descriptor_list;
		const struct descriptor_list_struct * e_end = e + DESCRIPTOR_LIST_ENTRIES;
		for( ; e != e_end; e++ )
		{
			if( e->lIndexValue == (uint32_t)USBFS_IndexValue )
			{
				ctx->pCtrlPayloadPtr = (uint8_t*)e->addr;
				len = e->length;
				break;
			}
		}
		if( e == e_end )
			goto sendstall;
		if( len > USBFS_SetupReqLen )
			len = USBFS_SetupReqLen;
		ctx->USBFS_SetupReqLen = len;

		break;
	}
```
So for the descriptors, it's filling the `ctx->pCtrlPayloadPtr` buffer. Why would it fill the `ctrl0buf` for `GET_STATUS`?

cnlohr:
These are questions we try to not ask ourselves.

biemster:
the GET_STATUS issue is addressed in #829, so this is ready to be merged now.

PR #817, opened 2025-12-27
fix x035 i2c sensor demo

biemster:
not sure how this got passed the CI, but these changes should fix it again. (I don't have a x035 to test it though)


cnlohr:
Thanks

PR #821, opened 2025-12-31
Refactor RTC code and add CH32L103 standby example

nolanhergert:


cnlohr:
This seems to be fine.  Is it ready for merging?

nolanhergert:
Was able to clean up most of the magic delays and remove unnecessary code. It works great now!

I would like to get some regression testing and overall good vibes from @unicab369 and @Outfisken if possible, since I moved/changed their examples for the other chips.

cnlohr:
I don't particularly see a reason not to merge it now, I'll  let it simmer for a day or two though, and hope they take a look.

biemster:
I tested the RTC examples on 582 and 570, they work fine there

nolanhergert:
I ran the original v30x rtc.c example on a CH32V203 (updated makefile and funconfig) and it doesn't run (I added blinks to confirm). I'm not sure where it is failing yet.

I flashed the new rtc.c example and it is running, but not executing the IRQHandler even after updating it to use RTCAlarm. I want to get it working, but it doesn't need to block this checkin in my opinion.

I'm commuting to CES soon, so I might not be able to get it working in a reasonable time frame.

nolanhergert:
Got the refactored code for the v30x example working, but it wasn't sure if the original code had a bug or not (software reset required RTC_init() again). Left some notes for the next person

cnlohr:
sounds good.

PR #823, opened 2026-01-02
Fixes to minichlink

monte-monte:
Fixes:
- Ardulink failing to initialize when MCU is powered from the GPIO. Added retries to setup function, ideally need a small fix in the ardulink firmware, but few retries couldn't hurt in general.
- Fixed nasty bug that was erasing wrong sectors on ch5xx in some cases. Fixed (hopefully) gdb on these chips.
- Fixed a part in minichgdb.c that assumed flash must always start at 0x800000. Not sure when it could break things and how, but this wouldn't work on ch5xx chips.
- Replaced unnecessarily complex function that read UUID on ch5xx chips. Was fooled by the official lib before.
- Added a flag for when running in gdb mode. This alters behavior of some functions on ch5xx (not changing clock, always checking memory partition).

### DIFF
--- a/minichlink/minichlink.c
+++ b/minichlink/minichlink.c
@@ -2017,6 +2017,7 @@ int DefaultWriteBinaryBlob( void * dev, uint32_t address_to_write, uint32_t blob
 	//  this is only fallback functionality for really realy basic programmers.
 	//  it is also used in unbrick.
 
+	int ret = 0;
 	uint32_t rw;
 	struct InternalState * iss = (struct InternalState*)(((struct ProgrammerStructBase*)dev)->internal);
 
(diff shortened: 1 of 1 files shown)


monte-monte:
Wait, don't merge. Fixing RAM writes...

monte-monte:
Done

monte-monte:
- Added checking returns of all write functions in ``DefaultWriteBlob`` to not return success on failed writes.
- Improved warning message about read protection being enabled.
Fixes https://github.com/cnlohr/ch32fun/issues/825

minichlink/ch5xx.c
```diff
@@ -1046,7 +1082,7 @@ int CH5xxWriteBinaryBlob(void * dev, uint32_t address_to_write, uint32_t blob_si
 				write_function = &ch5xx_write_flash;
 			}
 		} else {
-			if (blob_size > 4096) write_function = &ch5xx_write_flash_using_microblob2;
```
cnlohr, line 1049: Is this really what you meant to do?  Do we really want to use this for small blobs?

minichlink/ch5xx.c
```diff
@@ -1068,13 +1104,12 @@ int CH5xxWriteBinaryBlob(void * dev, uint32_t address_to_write, uint32_t blob_si
 			else new_blob_size = 0;
 		}
 
+		CH5xxErase(dev, address_to_write, blob_size, 0);
```
cnlohr, line 1107: Are you really sure about this too?  The issue the person was seeing wasn't that the data wasn't erased.

cnlohr: nvm I See below

minichlink/ch5xx.c
cnlohr: I won't block for this comment.

monte-monte: It's faster this way, and some people were complaining...

minichlink/ch5xx.c
monte-monte: Previous erase was just wrong it erased at wrong offset.

PR #827, opened 2026-01-11
Added SPI Slave/Peripheral example

UserR00T:
This PR adds an example SPI slave implementation. Whilst the implementation is quite barebones in terms that it doesn't support DMA, it works to accurately echo the first two bytes ot the MOSI line and two static bytes through the MISO line. This demo is mostly intended as to show how to configure the SPI interface as a slave device and what exactly is needed in terms of main loop and such. A great implementation of an SPI slave device was found in `spi_dma_chain`, but this follows the SPI chain setup and thus requires two SPI busses which I tried to implement as a standalone demo.

I'm not an low-level/embedded developer and don't work a lot with C (mostly work with higher level languages for my day-to-day work), let me know if I missed anything obvious (e.g. optimizations and/or linting)

Hope this demo helps!

### DIFF
--- a/examples/spi_slave/spi_slave.c
+++ b/examples/spi_slave/spi_slave.c
@@ -1,13 +1,6 @@
 #include "ch32fun.h"
 #include <stdio.h>
 
-#define CH32V003_SPI_SPEED_HZ 1000000
-#define CH32V003_SPI_DIRECTION_2LINE_TXRX
-#define CH32V003_SPI_CLK_MODE_POL0_PHA0
-#define CH32V003_SPI_NSS_SOFTWARE_ANY_MANUAL
-
-#include "../../extralibs/ch32v003_SPI.h"
-
 #define PIN_CS PC1
 // Useful to determine the cycle duration with a scope. Goes high the moment CS goes low, and back to low at the last
 // action within the while loop. Can be removed otherwise.
@@ -34,6 +27,16 @@ void wait_for_state( SpiState desired_state )
 	while ( spi_state != desired_state )
 	{
 	}
+}
+
+static uint8_t SPI_read_8()
+{
+	return SPI1->DATAR;
+}
+
+static void SPI_write_8( uint8_t data )
+{
+	SPI1->DATAR = data;
 }
 
 void SPI_Configure()


examples/spi_slave/spi_slave.c
```diff
@@ -0,0 +1,176 @@
...
+void SPI_Configure()
+{
+	// reset control register
+	SPI1->CTLR1 = 0;
+
+	// Enable GPIO Port C and SPI peripheral
+	RCC->APB2PCENR |= RCC_APB2Periph_GPIOC | RCC_APB2Periph_SPI1;
+
+	// PC5 is SCLK
+	GPIOC->CFGLR &= ~( 0xf << 4 * 5 );
```
cnlohr: Consider using 
```c
funPinMode( PC5, GPIO_CNF_IN_FLOATING );
```
and the GPIO AF etc.  

This compiles to the same code, but is easier for most to read.

cnlohr:
Only one comment - to switch to the new way to do `funPinMode`

cnlohr:
Please please update it for that, I really like this!  It will be really helpful.

UserR00T:
No worries, I've refactored the signature to the new method/definition. I was also thinking that it might be handy to call `SPI_init` from here and configure it as `SPI_Mode_Slave`, thoughts? Or should we instead keep the `ch32v003_SPI.h` implementation mostly master-specific apart from the RX/TX methods? I do think things like `CH32V003_SPI_DIRECTION_2LINE_TXRX` get more complex the moment we introduce the mode variable, as the MOSI/MISO lines need to be inverted depending on the current SPI mode (floating/pushpull) -- maybe it would even be beneficial to refactor this to a `ch32_SPI.h` or something, and make it platform/MCU agnostic by requiring the user to configure the pinout themselves. That way it becomes less of an issue, and personally I think these standardized extra libraries could be quite helpful in getting the groundwork setup for your own project; but I'm very interested in hearing your opinion about this.

I was thinking something like
```c
#include "../../extralibs/ch32v003_SPI.h"

#define CH32V003_SPI_MODE SPI_Mode_Slave // <-- this would be new
#define CH32V003_SPI_SPEED_HZ 1000000
#define CH32V003_SPI_DIRECTION_2LINE_TXRX
#define CH32V003_SPI_CLK_MODE_POL0_PHA0
#define CH32V003_SPI_NSS_SOFTWARE_ANY_MANUAL

int main()
{
	SystemInit();
    SPI_init();
}
```

Either way the PR is ready to merge now, let me know!

examples/spi_slave/spi_slave.c
```diff
@@ -1,4 +1,4 @@
-#include "ch32fun.h"
+#include "../../ch32fun/ch32fun.h"
```
cnlohr: why did this change?  -I should handle this.

cnlohr:
Longer-term it may make sense to do a larger refactor, for a separate example.  But, it's important to make examples that work independent of any extra libraries or abstractions, and this does that.

But also, plz don't use relative paths.

UserR00T:
Alright, that makes sense. My bad about the relative path, I did revert that already in cf59678. I think I misconfigured clion as compiling works great, but from within the ide it wouldn't work without an relative path; which change I forgot to unstage before commiting.



15 Jan 2026, 09:21 by ***@***.***:

cnlohr:
Mind fixing it now?  I unblocked for CI testing.

UserR00T:
I fixed the CI errors, seems to compile now (had to define some "dummy" values even when unused)

edit: Now that I'm looking over it again, it may even be alright to locally define the `SPI_write_8`/`SPI_read_8` sections as that's basically the only thing I need from `ch32v003_SPI.h`; would save unnecessary unused definitions & imports. Let me know, happy to change it if that's better

examples/spi_slave/spi_slave.c
```diff
@@ -0,0 +1,176 @@
+#include "ch32fun.h"
+#include <stdio.h>
+
+#define CH32V003_SPI_SPEED_HZ 1000000
+#define CH32V003_SPI_DIRECTION_2LINE_TXRX
+#define CH32V003_SPI_CLK_MODE_POL0_PHA0
+#define CH32V003_SPI_NSS_SOFTWARE_ANY_MANUAL
+
+#include "../../extralibs/ch32v003_SPI.h"
```
cnlohr: Please just use "ch32v003_SPI.h"

cnlohr: Or not include it at all.

cnlohr:
I'd honestly you remove the reference to ch32v003_SPI.h.  And just include the content locally.

UserR00T:
I agree that's better. I've just done that change and the PR is once again ready for review. At last it should be totally fine now haha:tm:

cnlohr (commented):
Looks good.

cnlohr:
oh crud, I guess my review from before didn't submit.  But you fixed it anyway, so all is well.

PR #828, opened 2026-01-12
fixed makefile path

eeucalyptus:
There seems to be a mistake with the location of the makefile. 

Fixes the build error 

```
Makefile:17: ../../../ch32fun/ch32fun.mk: No such file or directory
```


cnlohr:
thanks

PR #829, opened 2026-01-12
Bug fixes

monte-monte:
- fixes GET_STATUS request not working in fsusb.c
- minor addition to send ep function in hsusb.c
- enable ch32v317 support in minichlink
- add special procedure for bootloader flashing on ch32v20x/v30x
- fix flash write error on ch5xx (timing of PC running minichlink was an issue)


Grazfather:
Confirmed that the bootloader flashing procedure worked for me!

cnlohr:
Lots of complicated stuff in there, but if it's good, it's good.

PR #832, opened 2026-01-17
USBFS usbprintf input handler

biemster:
The debug printf has a `handle_debug_input` that is fed by `poll_input()`, this change package adds the same api for usbprintf, with the handler called `handle_usbfs_input` and the poller keeps the name `poll_input()`.

The `USBFS/usbfs_cdc_tty` demo is updated to show how to use it, the micropython project needs it to provide REPL over USB as well.


cnlohr:
Looks good.

PR #837, opened 2026-01-20
Add 64-bit safe timer functions for 32- and 64-bit timers

cnlohr:


maxgerhardt:
This PR has placed the function declaration

```cpp
uint64_t funSysTick64( void );
```
outside the `#ifndef __ASSEMBLER__` block, meaning assembly files that `#include <ch32fun.h>` see this declaration and fail to build. This destroyed rv003usb's ability to compile against the latest ch32fun version.

PR #843, opened 2026-01-26
Fix EP5-7 in fsusb

monte-monte:
- Adding missing definitions for EP5-EP7 to ``fsusb.h``
- Fixed ``USBFS_TypeDef`` for ch5xx
- Added ability for directly setting EP0 buffer in custom setup handler, to prevent extra copy.
- Changed ``usbfs_i2c.c`` to use this method.

### DIFF
--- a/extralibs/fsusb.h
+++ b/extralibs/fsusb.h
@@ -29,9 +29,9 @@ typedef struct
 	__IO uint8_t  MIS_ST;
 	__IO uint8_t  INT_FG; // "Combined" register in some situations. (ST_FG)
 	__IO uint8_t  INT_ST;
-	__IO uint8_t RX_LEN;
+	__IO uint8_t  RX_LEN;
 	__IO uint8_t  Reserved0;
-	__IO uint16_t  Reserved1;
+	__IO uint16_t Reserved1;
 	__IO uint8_t  UEP4_1_MOD;
 	__IO uint8_t  UEP2_3_MOD;
 	__IO uint8_t  UEP567_MOD;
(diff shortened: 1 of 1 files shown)


monte-monte:
DMA buffers for EP > 3 are very inconvenient on ch5xx and ch32x035. First there is a huge gap between EP3 DMA and EP5 DMA regs, so you can't use a macro effectively (and I was unable to use conditional ternary notation for this).
Then EP4 DMA can't be set on it's own. It always comes at 64 bytes offset after EP0 DMA buffer address. No idea how to integrate EP4 in the lib at the moment, so those who wants to use it with these MCUs would need to setup it manually.

biemster:
> First there is a huge gap between EP3 DMA and EP5 DMA regs, so you can't use a macro effectively (and I was unable to use conditional ternary notation for this).

Macros don't support ternary indeed, but you can ternary the index like this:
```C
[n*2 +((n>4) ? (&USBFS->UEP5_DMA - &USBFS->UEP0_DMA -10) : 0)]
```
EDIT: almost! but you get the idea

Also you missed a field at the end of the ch5xx struct:
image

image

biemster:
I can add this in the USBFS Mass Storage PR that is coming this week, and I would prefer the ternary in `UEP_DMA` which would mean reverting the changes to `fsusb.c`, but we could leave it like this change as well if others prefer.

monte-monte:
@biemster yeah, missed that one, sorry. I wonder how it worked in my tests...

I wanted to do ternary the other way and it didn't work. Yours seems to work, but it will also try to do this for EP4 and then it will return ``UEP0_TX_LEN``. Maybe checking EP number is better after all?

biemster:
Also the `UEP_CTRL_*` macros for ep 5,6,7 need this treatment, not only the DMA macro?

PR #844, opened 2026-01-27
USBFS Mass Storage

biemster:
USBFS Mass Storage demo using a small RAM disk to store an INDEX.HTM with a link to the github repository.
This could easily be extended to store the file in flash with the new ch5xx flash library, although that would make it ch5xx only. Also this demo is quite large as it is, it might have to move to `projects/` then.

### DIFF
--- a/examples_usb/USBFS/usbfs_msc/Makefile
+++ b/examples_usb/USBFS/usbfs_msc/Makefile
@@ -1,8 +1,8 @@
 all : flash
 
 TARGET:=usbfs_msc
-TARGET_MCU:=CH585
-TARGET_MCU_PACKAGE:=CH585M
+TARGET_MCU:=CH570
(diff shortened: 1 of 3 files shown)


biemster:
~~I did not implement writing to the file yet, after that and~~ if we don't want it persisted to flash this PR is ready.
EDIT: turns out I _did_ implement writing to the file already :muscle: , so only flash persistence is up for discussion.

biemster:
wait no, writing might not be working after all. It's possible linux just wrote to the cached version but never flushed.

biemster:
I converted to draft again, because I almost got writing to work.

biemster:
Writing from PC to MCU works now, this demo is a lot of fun!

cnlohr:
All looks good.

PR #847, opened 2026-01-31
remove debugging log that caused a warning

eeucalyptus:
This fixes a compiler warning in the common 64-bit systick


eeucalyptus:
Should the "Trigger" log also be removed?

cnlohr:
yes and fixed it too directly in master.

PR #848, opened 2026-02-01
iSLER: define all struct fields as {BB,LL,RF}.{BB,LL,RF}xx

biemster:
This small change makes all fields in the BB, LL and RF structs a `uint32_t` field named {BB,LL,RF}xx, to clarify where they live when reversing them. The following fields were common to all BLE enabled WCH chips so they had a specific name in the structs:
```C
#define CTRL_CFG           BB0
#define CRCINIT1           BB1
#define STATUS             LL2
#define INT_EN             LL3
#define CTRL_MOD           LL20
#define TXTUNE_CTRL        RF14
#define TXCTUNE_CO_CTRL    RF36
#define TXCTUNE_GA_CTRL    RF37
#define RXTUNE             RF39
#define TXCTUNE_CO         RF40
#define TXCTUNE_GA         RF50
```
They got this new macro so nothing will change in the API or anywhere else, this is just to help in the ongoing RE effort.

### DIFF
--- a/extralibs/iSLER.h
+++ b/extralibs/iSLER.h
@@ -44,6 +44,20 @@
 #include <stdio.h>
 #include <string.h>
 
+// common fields
+#define CTRL_CFG           BB0
+#define CRCINIT1           BB1
+#define STATUS             LL2
+#define INT_EN             LL3
+#define CTRL_MOD           LL20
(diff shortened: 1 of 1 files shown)


cnlohr:
spectacular

PR #849, opened 2026-02-01
Add USBD extralib, CDC TTY usbprintf demo

biemster:
Several chips that have hw USB also have the USBD peripheral. ch32fun already has an extralib for USBFS, which is available on all hw USB enabled chips, but some also have this USBD peripheral. While USBFS is preferred since it has more functionality, some dev boards only have the USBD pins broken out on pads.
One very prominent example is the ch32v208 boards on ali with Ethernet for less than $3.

This package adds support for this USBD peripheral, and an `usbprintf` demo using CDC TTY.

The code in the USBD extralib is heavily inspired by R's work in https://github.com/ArcaneNibble/wch-uf2.


biemster:
works now! :partying_face:  Tested on v208

cnlohr:
This looks pretty good. I don't have any comments on it.  Thank you for this.

PR #852, opened 2026-02-06
iSLER: make highcode section configurable

biemster:
This change fixes compilation of iSLER on chips without the __HIGH_CODE macro (`v208`), and makes the section configurable by adding something like
```C
#define FUNCONF_ISLER_SECTION __attribute__( ( section( ".manual_ram" ) ) )
```
in the project's `funconfig.h`.


cnlohr:
Looks good.

PR #853, opened 2026-02-07
USBD: share pma buffer if endpoint is bidirectional

biemster:
A small update already from real world experience on #849 ; the packet sram for USBD is very limited, and can hold only 6 endpoint buffers. Endpoint 0 is bidirectional, and if you want CDC ACM (3 endpoints) and one more two-way interface that's already 7 endpoint directions, so wouldn't fit. This change allows bidirectional endpoints to share the packet buffer (which is only endpoint 0 for now, full bidirectional on other endpoints is not supported yet).

This reduces the number of buffers needed for a dual interface connection including CDC ACM to 6, which exactly fits in the packet SRAM.
This is tested working on the MicroPython port on v208.

(there is also some clarifications on comments on the USBD code, and a more robust clock setup copying what was already implemented for USBFS)

### DIFF
--- a/extralibs/usbd.c
+++ b/extralibs/usbd.c
@@ -401,27 +401,32 @@ void USBD_InternalFinishSetup() {
 	USBDCTX.endpoint_mode[7] = FUSB_EP7_MODE;
 #endif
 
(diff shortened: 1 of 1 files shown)


cnlohr:
this is great.

Issue #855, opened 2026-02-09
Compiler generates recursive memset call


sellicott:
From a discussion that @monte-monte and I had on the discord yesterday.
The implementation of `memset` in ch32fun will produce recursive calls at certain optimization levels.
For reference, see https://godbolt.org/z/WqG86ocn8

This is because gcc expects the standard library to be available, so that it can optimize loops to library functions (memcpy, strcpy, memset, etc)
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56888

The solution seems to be to add the -ffreestanding option to the CFLAGS to disable this behavior.

PR #856, opened 2026-02-09
Add -fno-tree-loop-distribute-patterns to cflags

sellicott:
Attempt to fix https://github.com/cnlohr/ch32fun/issues/855

I haven't tested this, so I would request that @monte-monte and or @cnlohr checks before merging.


cnlohr:
I am going to trust you on this one, so far you haven't steered me wrong.

PR #857, opened 2026-02-10
Add Coremark to projects

monte-monte:
Adding a Coremark port + build script. Running benchmark is as easy as choosing your MCU type and typing ``make``.

### DIFF
--- a/projects/coremark/Makefile
+++ b/projects/coremark/Makefile
@@ -1,0 +1,53 @@
+all : flash
+
+TARGET:=coremark
+# TARGET_MCU:=CH32V006
+# TARGET_MCU:=CH32X035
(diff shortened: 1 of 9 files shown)


cnlohr:
@monte-monte  Because ch32fun is frequently used as a submodule itself, it should not contain submodules, for coremark, please make `make` check out a copy of `coremark` locally using svn commands, instead of this being stored as a submodule, other than that this looks great.

monte-monte:
I'll just leave this to the user then. Not a big deal.

cnlohr:
I mean please make a `make` target to clone it.

cnlohr:
@monte-monte if you can still please add an easy to way to get-and-compile this via the makefile.

PR #868, opened 2026-02-28
Fix horizontal byte order and bit-order handling in ssd1306_drawImage()

serebent:
I’ve updated the following:
- Fixed the 8‑pixel shift right issue
- Fixed the left–right mirroring issue
- Added support for both LSB‑first and MSB‑first bitmap data

If you define `"#define SSD1306_BMP_LSB_FIRST"` before `"#include ssd1306.h"`, the driver will treat the bitmap as LSB‑first.If nothing is defined, it defaults to MSB‑first as before.

I ran some simple tests.
- https://github.com/serebent/ch32fun-test


cnlohr:
Thanks for putting in the work! I really appreciate it.

cnlohr:
This was a lot of work, including the tests. I applaud your efforts.  All tests passed, merging now.

PR #878, opened 2026-03-26
Fix minichlink can't work when combination of some options

YuukiUmeta-UIAP:
- https://github.com/cnlohr/rv003usb/issues/128 has been fixed
- Other target programmers could also have same bugs like the above URL.


cnlohr:
@monte-monte can you review this?

cnlohr:
(it looks good to me)

monte-monte:
@cnlohr if you're okay with that, I have no objections either. Except that buffer size could be bigger, if the targeted programmer uses new firmware. But I can change it later, when I test new boards.

PR #883, opened 2026-04-24
Expand IRQ handling for CH32x00x models and add dual USART example

stefkuhb:
Added missing entries for USART2 and OPCM used by the ch32x00x models


cnlohr:
Looks good, will merge on CI completion.

PR #884, opened 2026-04-26
minichlink: fix missing ch584 detection

karlp:
CH584 was being detected as ch585, as the and masking makes them "match" because it didn't check the more explicit 584 from the list first.


cnlohr:
Looks good!

PR #885, opened 2026-05-06
Changes to minichlink and ssd1306.h

monte-monte:
- Updated binary stubs for rv003usb bootloaders
- Added support for variable buffer size (on windows can't use sizes over 3200 byte, need to find out why)
- Added support for new v00x bootloader
- Added warning about flashing with ISP on ch5xx when debug module is enabled
- Added recognition of CH584 chip
- Added versioning of minichlink builds. From now on if you push an update to minichlink please do ``make hash`` in minichlink folder, so we can find easily versions in commit history.
- Added 3-wire SPI mode for SSD1306(and other compatible) displays.


cnlohr:
Is this ready for merging? The changes look reasonable.

cnlohr:
Looks good.

PR #887, opened 2026-05-07
CH5xx WS2812B by PWM/TMR/DMA/UART example

mattshepcar:
An example of some other ways to drive those fancy addressable LEDs on CH5xx MCUs for when you don't want to use the SPI pins or you need even more LEDs.


cnlohr:
@mattshepcar why not just use the code in the WS2812B driver in extralibs? https://github.com/cnlohr/ch32fun/blob/master/extralibs/ws2812b_dma_spi_led_driver.h#L243-L273 it works on the CH5xx

mattshepcar:
Mainly because I don't want to or can't use the SPI pins.  This also gives an example of how to set up PWM/TMR DMAs which could be used for things other than driving WS2812Bs.

examples_ch5xx/ws2812b_pwm/ws2812b_pwm.c
```diff
@@ -0,0 +1,231 @@
...
+	R8_TMR2_CTRL_MOD = RB_TMR_OUT_EN|RB_TMR_COUNT_EN;
+	R8_TMR2_CTRL_DMA = RB_TMR_DMA_ENABLE;
+}
+#endif
+
+#ifdef CH57x
+#define TMR_PA9 1
+// the PWM buffers start and end with zero entry (duty cycle of 0 = output off). 
+// I'm not sure why the start one is needed but you get flickering without it.
+uint16_t __attribute__((aligned(4))) WS2812B_PWM123_LedBuffer[(NR_LEDS * 24 + 2) * 3] = {0};
```
cnlohr, line 119: Can you add a comment here explaining why the alignment?  (I.e. because these are the raw buffers that are sent out via DMA)

examples_ch5xx/ws2812b_pwm/ws2812b_pwm.c
```diff
@@ -0,0 +1,231 @@
...
+#endif
+// the TMR buffer just needs one zero at the end to turn off the output
+uint32_t WS2812B_TMR_LedBuffer[NR_LEDS * 24 + 1] = {0};
+
+#ifdef RB_PIN_U0_INV // CH584/5 support inverted UART output
+uint32_t WS2812B_UART_LedBuffer[NR_LEDS];
+
+static const uint8_t WS2812B_UART_LUT[8] = 
+{
+	~0b0100100, // 9 bits are used to encode 3 bits of WS2812B data.
```
cnlohr, line 130: You don't need to change this! But, specifying ~ in a LUT is not normally a natural way to express this, instead, you would normally write out the inverted bitstream in the LUT.  This is to make debugging easier.

examples_ch5xx/ws2812b_pwm/ws2812b_pwm.c
```diff
@@ -0,0 +1,231 @@
...
+	~0b1100100, // Each of these 7 bit payloads is framed by the
+	~0b0101100, // UART high-start and low-stop bits. Entries are
+	~0b1101100, // inverted to compensate for the output inversion.
+	~0b0100101,
+	~0b1100101,
+	~0b0101101,
+	~0b1101101,
+};
+void WS2812B_UART0_Send(const uint32_t* ledData, int numLeds)
```
cnlohr, line 139: Can you write a blirb about these functions and what they do, and whether the ledData pointer is used as the DMA pointer, and/or if the data is made available upon completion or after completion.

examples_ch5xx/ws2812b_pwm/ws2812b_pwm.c
```diff
@@ -0,0 +1,231 @@
...
+{
+	uint32_t* dst = &buffer[pixelNumber * 24];
+	rgb <<= 8;
+	for(int bit = 0; bit != 24; ++bit, rgb <<= 1)
+	{
+		*dst++ = WS2812B_0 + ((rgb >> 31) & (WS2812B_1 - WS2812B_0));
+	}
+}
+#ifdef CH57x
+void WS2812B_PWM_Init(int channel)
```
cnlohr, line 24: Normally, you would make these zero indexed, even though it would disagree with the datasheet.  So because you are matching the datasheet, can you change it around so that makes it clear that it's pwm_channel_1_indexed

cnlohr:
this is really close, do you mind fixing the things I outlined?  they're mostly documentation.

cnlohr:
@mattshepcar up for addressing the comments?

PR #889, opened 2026-05-09
additions to make PortE work on ch32v307VCT6

JKTsDoodles:
These additions were necessary to make the function funGpioInitAll() and funDigitalWrite() to work as expected for PortE on a board with a ch32v307VCT6. I've tested funDigitalWrite() for each pin on port E and verified the expected operation in hardware with a scope.


cnlohr:
Looks reasonable to me, any chance I could get a gander by @monte-monte in case they know of any chips where the port E bit would break something on some chips.

monte-monte:
PE port bit is only absent in v103, it is *reserved* there. But I think writing it shouldn't cause a problem (it is described as RO in the datasheet).

PR #890, opened 2026-05-13
Add SH1106 display mode to ssh1306.h

monte-monte:
This adds a missing mode for 1.5 inch SH1106 128x64 oled displays. It fixes offset setting for SH1107 display driver part of the code.


cnlohr:
Looks good to me!

PR #893, opened 2026-05-27
Optimize MakeGDBPrintText by caching strlen result

rozanlaudzai:
In `minichlink/microgdbstub.h`, `strlen(msg)` was being evaluated on every iteration of the loop.

By caching the length in `msg_len` at the start of the function, we reduce the time complexity of `MakeGDBPrintText` from O(N^2) to O(N).

### DIFF
--- a/minichlink/microgdbstub.h
+++ b/minichlink/microgdbstub.h
@@ -186,13 +186,14 @@ void SendReplyFull( const char * replyMessage )
 }
 
 void MakeGDBPrintText(const char* msg) {
+	size_t msg_len = strlen(msg);
 	// ASCII to hex conversion doubles size, plus 'O', plus NUL
-	size_t buf_len = 2 * strlen(msg) + 2;
+	size_t buf_len = 2 * msg_len + 2;
 	char* buf = alloca(buf_len);
 	memset(buf, 0, buf_len);
 	buf[0] = 'O'; // for "Output"
 	char* target = buf + 1; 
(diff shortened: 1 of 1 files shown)


cnlohr:
Such a strange PR.  I've been getting a deluge of PRs from AI agents about really dumb things, and this one almost fell through the cracks.

But this is legit.  Thanks for finding this.

Nable80:
IMHO, if there's a non-zero chance for a possible situation when some compiler won't optimize this classical construction on its own, I guess it makes sense to replace ``memset(buf, 0, buf_len);`` with a single write too, e.g. ``buf[buf_len - 1] = '\0';``

As someone who wrote too much code in assembly, I'd even prefer to rewrite it like that:
```c
	buf[0] = 'O'; // for "Output"
	char* target = buf + 1;
	for( size_t i = 0; i < msg_len; i++ ) {
		*target++ = ToHEXNibble((msg[i] & 0xf0u) >> 4u);
		*target++ = ToHEXNibble(msg[i] & 0x0fu);
	}
	*target = '\0';    
```
or even:
```c
	char* target = buf;
	*target++ = 'O'; // for "Output"
	for( size_t i = 0; i < msg_len; i++ ) {
		*target++ = ToHEXNibble((msg[i] & 0xf0u) >> 4u);
		*target++ = ToHEXNibble(msg[i] & 0x0fu);
	}
	*target = '\0';    
```

There's no chance for vectorization here anyway, so this dependency between two lines shouldn't hurt.

PR #895, opened 2026-05-30
fix -Wunused-parameter warnings

20dka:
fixes for `[-Wunused-parameter]`

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -1451,6 +1451,7 @@ void handle_reset( void )
 
 __attribute__ ((naked)) int setjmp( jmp_buf env )
 {
+	(void)env;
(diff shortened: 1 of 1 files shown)


maxgerhardt:
This reminds me that the asm implementation doesn't seem to be LTO safe, just like I discovered it with a bug in https://github.com/Community-PIO-CH32V/framework-wch-noneos-sdk/issues/14. Surpressing the warning about the unused parameter is the wrong way, it needs to be referenced in the assembly to make sure the first parameter is absolutely in the expected a0 register.

cnlohr:
We talked about this on-call.  I think @20dka is going to be tweaking this.

cnlohr:
@maxgerhardt do you mind looking into this so we  can close this out?

PR #899, opened 2026-06-01
Fix flash size in gdb. Add a warning for ch570 ISP

### LINKED ISSUE #879: Bug Report: GDB Memory Map Unit Mismatch (Flash size in KB vs Bytes) on CH32V003
### **1. The Problem (Symptom)**
When using the latest `ch32fun` (master branch) with a CH32V003, the `next` and `step` commands fail immediately upon starting a debug session. GDB reports that it cannot access memory at the current PC, even though the address is valid Flash memory. This effectively prevents source-level stepping in modern GDB clients.

**Test Code used for reproduction:**
```c
#include "ch32fun.h"
#define TIME_DELAY_MS 100

void GPIO_PC4_Init() {
    RCC->APB2PCENR |= RCC_APB2Periph_GPIOC;
    GPIOC->CFGLR &= ~( 0xf << ( 4 * 4 ) ); 
    GPIOC->CFGLR |= ( 0x1 ) << ( 4 * 4 ); 
}

int main() {
    SystemInit();
    GPIO_PC4_Init(); // <--- GDB fails here during 'next'
    while ( 1 ) {
        GPIOC->BSHR = ( 1 << 4 );
        Delay_Ms( TIME_DELAY_MS );
        GPIOC->BSHR = ( 1 << 20 );
        Delay_Ms( TIME_DELAY_MS );
    }
}
```

**GDB Error Output:**
```text
Breakpoint 1, main () at blink_raw.c:18
18         SystemInit();
(gdb) next
Cannot access memory at address 0x20c6
```

---

### **2. Investigation (The GDB Trace)**
By enabling remote debugging logs (`set debug remote 1`), we identified that the GDB server (via `microgdbstub`) is misreporting the chip's memory boundaries during the initial XML handshake.

**Observed GDB Packet:**
```text
[remote] Sending packet: $qXfer:memory-map:read::0,1000#ab
[remote] Packet received: l<?xml version="1.0"?>...
<memory type="flash" start="0x00000000" length="0x10">
```
The server reports `length="0x10"` (16 bytes). Modern GDB clients (tested with GDB 12.1+ on Ubuntu 22.04) strictly enforce this map. Because the Program Counter `0x20c6` is significantly outside that 16-byte range, GDB assumes the address is invalid/unmapped and aborts the command without attempting to query the hardware.

---

### **3. Root Cause**
There appears to be a unit mismatch or a variable truncation occurring during the toolchain's initialization. While `chips.c` correctly defines the CH32V003 flash size in bytes:

```c
// chips.c
.flash_size = 16*1024, // 16384 bytes
```

The value that eventually reaches the `InternalState` (`iss`) used by the GDB stub is being treated as if it were in Kilobytes (value `16`). When this value is formatted into the XML string using `%x`, it results in the `0x10` (16 byte) limit seen by GDB.

**File:** `microgdbstub.h`
**Line:** ~307
```c
// Current logic: iss->flash_size is passed directly. 
// If the value is 16, GDB interprets this as 16 bytes.
snprintf( map, mslen, MICROGDBSTUB_MEMORY_MAP, iss->flash_size, iss->sector_size, iss->ram_size );
```

---

### **4. The Solution**
Ensuring the values are explicitly scaled to bytes before the XML generation resolves the issue. After applying the following fix and rebuilding `minichlink`, GDB correctly receives `length="0x4000"` (16,384 bytes) and debugging functions normally.

**Corrected Line:**
```c
snprintf( map, mslen, MICROGDBSTUB_MEMORY_MAP, iss->flash_size * 1024, iss->sector_size, iss->ram_size );
```
*(Note: If `iss->flash_size` is intended to be in bytes globally, a deeper investigation into why it is being truncated to 16 for the V003 may be required.)*

---

Given the transition to the new multi-chip architecture, I suggest a broader check of the `InternalState` struct initialization. It seems the abstraction between the `RiscVChip_s` definitions and the GDB stub may have introduced a "KB vs Bytes" inconsistency. A deeper audit would ensure that larger chips (V203, V307) do not suffer similar boundary truncation on strict GDB clients.

Finally, I want to say a huge thank you for your time and dedication to maintaining `ch32fun`. It is an incredible project that has made RISC-V development accessible to so many of us. I hope you find this report helpful, and if you have the time, I'd appreciate a quick investigation to ensure this unit mismatch hasn't crept into other parts of the new toolchain architecture. Thanks again for all the hard work!

monte-monte:
- Fixes https://github.com/cnlohr/ch32fun/issues/879
- Added a warning when trying to flash ch570/2 in ISP mode, but when read protection is disabled.
- Bumped exe and *version*


monte-monte:
Please don't merge, found one more bug.

cnlohr:
will not merge

monte-monte:
Added fixes for rv003usb bootloader. That affect the new v006 bootloader. It now disables interrupts befure writing flash and reenables them before exiting function.
Fixed and simplified the logic of selecting the size of the buffer/id of HID.

minichlink/pgm-b003fun.c
```diff
@@ -146,16 +146,42 @@ static const unsigned char run_app_blob[] = {
...
+// 	0x33, 0x86, 0x55, 0x00, 0xb7, 0x06, 0x09, 0x00, 0xd4, 0xc3, 0x94, 0x41,
+// 	0x54, 0x47, 0x94, 0xc1, 0xb7, 0x06, 0x05, 0x00, 0xd4, 0xc3, 0x94, 0x41,
+// 	0x91, 0x05, 0x11, 0x07, 0xe3, 0xc8, 0xc5, 0xfe, 0xb7, 0x06, 0x01, 0x00,
+// 	0x93, 0x86, 0x06, 0x04, 0xd4, 0xc3, 0x83, 0xa6, 0xc5, 0xff, 0xe3, 0xc9,
+// 	0x65, 0xfc, 0xfd, 0x56, 0x14, 0xc1, 0x82, 0x80
+// };
+
+// Interrupts disabled, no FLASH->ADDR loads
 unsigned char write_block_bin[] = {
-	0x13, 0x07, 0x45, 0x05, 0x0c, 0x43, 0x5c, 0x43, 0x83, 0x12, 0x87, 0x00,
```
cnlohr, line 150: Please include the assembly in a comment from whence this is generated.

emolitor:
Just noting ❤️ for this as I was banging my head on the CH570 until I noticed this caught my exact issue. 🤦  Thank you!

PR #903, opened 2026-06-03
Use serial number to choose WCH-LinkE when multiple ones plugged

DeqingSun:
This commit added -l argument to pass in a serial number. If serial number is passed, the pgm-wch-linke.c will use the WCH-LinkE with matching serial number. Rather than last one wins.

### DIFF
--- a/minichlink/pgm-wch-linke.c
+++ b/minichlink/pgm-wch-linke.c
@@ -106,9 +106,9 @@ static void wch_link_multicommands( libusb_device_handle * devh, int nrcommands,
 	va_end( argp );
 }
 
-#define WCH_LINKE_SERIAL_MAX 256
+#define USB_SERIAL_MAX 256
 
-static int wch_linke_read_serial( libusb_device *device, char *buf, size_t buflen )
+static int get_usb_serialnumber( libusb_device *device, char *buf, size_t buflen )
 {
 	struct libusb_device_descriptor desc;
 	if( libusb_get_device_descriptor( device, &desc ) != 0 || desc.iSerialNumber == 0 )
@@ -125,8 +125,8 @@ static int wch_linke_device_matches_serial( libusb_device *device, const char *w
 {
 	if( !want_serial || !want_serial[0] )
 		return 1;
-	char buf[WCH_LINKE_SERIAL_MAX];
-	if( wch_linke_read_serial( device, buf, sizeof( buf ) ) != 0 )
+	char buf[USB_SERIAL_MAX];
+	if( get_usb_serialnumber( device, buf, sizeof( buf ) ) != 0 )
 		return 0;
 	return strcmp( buf, want_serial ) == 0;
 }
@@ -178,8 +178,8 @@ static inline libusb_device_handle * wch_link_base_setup( int inhibit_startup, c
 			struct libusb_device_descriptor desc;
 			if( libusb_get_device_descriptor(device,&desc) != 0 ) continue;
 			if( desc.idVendor != 0x1a86 || desc.idProduct != 0x8010 ) continue;
-			char buf[WCH_LINKE_SERIAL_MAX];
-			if( wch_linke_read_serial( device, buf, sizeof( buf ) ) == 0 )
+			char buf[USB_SERIAL_MAX];
+			if( get_usb_serialnumber( device, buf, sizeof( buf ) ) == 0 )
 				fprintf( stderr, " (available: '%s')", buf );
 		}
 		fprintf( stderr, "\n" );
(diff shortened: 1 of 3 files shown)


minichlink/pgm-wch-linke.c
```diff
@@ -106,7 +106,32 @@ static void wch_link_multicommands( libusb_device_handle * devh, int nrcommands,
 	va_end( argp );
 }
 
-static inline libusb_device_handle * wch_link_base_setup( int inhibit_startup )
+#define WCH_LINKE_SERIAL_MAX 256
+
+static int wch_linke_read_serial( libusb_device *device, char *buf, size_t buflen )
```
maxgerhardt: This function may be better called `get_usb_serialnumber()`, as it's generic and also called in the case of ARM and IAP programmer below. Same for `WCH_LINKE_SERIAL_MAX` macro above.

cnlohr:
I'd like @monte-monte 's comments on this as well.

monte-monte:
That's good, but I would put le_serial to a linke device struct.
This way we can add same functionality to other programmers in the future.

monte-monte:
I guess, what I meant to say is to describe it in minichlink.c as *programmer_serial* not wch_linke_serial. Otherwise if it works it works, gj.

cnlohr:
I agree, that would be really cool. @DeqingSun Are you up for making the change from references directly to Link-E and making it more generic. This will matter more as there are starting to be more CH32-based programmers.

minichlink/pgm-wch-linke.c
```diff
@@ -106,7 +106,32 @@ static void wch_link_multicommands( libusb_device_handle * devh, int nrcommands,
...
+		return -1;
+	libusb_device_handle *handle;
+	if( libusb_open( device, &handle ) != 0 )
+		return -1;
+	int r = libusb_get_string_descriptor_ascii( handle, desc.iSerialNumber, (unsigned char *)buf, buflen );
+	libusb_close( handle );
+	return ( r >= 0 ) ? 0 : -1;
+}
+
+static int wch_linke_device_matches_serial( libusb_device *device, const char *want_serial )
```
maxgerhardt: This is again a generic function that should be called `usb_device_matches_serial()`.

DeqingSun:
OK, now the serial number match is more generic.
There are 3 programmers using libusb to operate the programmer directly:
1. LinkE, this works
2. WCHISP, the WCH bootloader does not provice serial number
3. NHC-Link042, this does not have searching code, only ```libusb_open_device_with_vid_pid(ctx, 0x1986, 0x0034);```. I do not want to touch the code without hardware to test.

cnlohr:
@maxgerhardt @monte-monte thumbs up?

minichlink/pgm-wch-linke.c
```diff
@@ -123,16 +148,45 @@ static inline libusb_device_handle * wch_link_base_setup( int inhibit_startup )
...
+			if( usb_device_matches_serial( device, want_serial ) )
+				found_arm_programmer = device;
+		}
+		if( r == 0 && desc.idVendor == 0x4348 && desc.idProduct == 0x55e0) {
+			if( usb_device_matches_serial( device, want_serial ) )
+				found_programmer_in_iap = device;
+		}
 	}
 
+	if( filter_serial && !found && saw_linke )
```
maxgerhardt, line 175: Theoretically, this is missing the the cases for the other programmers. Throwing a "Proggammer x was seen but no matching serial yy found, available: xx" error for them might be helpful, too. That is, the ARM and IAP programmer. The code below can be generalized.

DeqingSun: As I said, the WCHISP, the WCH bootloader does not provide a serial number. So this can not be implemented.
I can try to add the case for other programmers but I can not validate the code because I do not have those programmers.

maxgerhardt: I mean, in the code above the ARM programmer is checked with usb_device_matches_serial() and equally the `desc.idVendor == 0x4348 && desc.idProduct == 0x55e0` programmer, so if we check for a matching serial there then we should print an equal error if we saw *a* device connected but none matching our serial.

DeqingSun: Sorry I misunstood, I will add it later today.

DeqingSun: Here is a sample output from the new commit

```
./minichlink -C linke -l whateverSerial
minichlink version - 2e47ba76b6165de99538d7759e48fb95aef89bb7
No matching WCH-LinkE (serial 'whateverSerial') found. (available RISC-V: '894E8F060E2C')
Error: Could not initialize linke programmer
```

cnlohr: that looks great.

cnlohr:
Thanks for pushing this to the end.  This looks great.

monte-monte:
This somehow breaks minichlink in windows.

DeqingSun:
> This somehow breaks minichlink in windows.

It compiles on my Windows machine with WSL. I'll check the WCH linke hardware connection later.

monte-monte:
It compiles, it doesn't work. I fixed it here https://github.com/cnlohr/ch32fun/pull/934

PR #904, opened 2026-06-04
Support 192k ram split

DeqingSun:
The new version of reference manual seems use 3 bits instead of 2 bits for Ram split.

Screenshot 2026-06-03 at 2 38 37 AM

So the split_code is changed to 3 bits.

Also fixed the bug that the chip became Read Protected after ram split, by restore RDPR properly.

### DIFF
--- a/minichlink/minichlink.c
+++ b/minichlink/minichlink.c
@@ -2863,7 +2863,9 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 	flash_ctlr &= CR_OPTER_Reset;
 	flash_ctlr |= CR_OPTPG_Set;
 	if( MCF.WriteWord( dev, (intptr_t)&FLASH->CTLR, flash_ctlr ) ) goto flashoperr;
-	if( MCF.WriteWord( dev, (intptr_t)&OB->RDPR, RDP_Key ) ) goto flashoperr;
+	uint16_t rdpr_bytes = (uint16_t)RDP_Key;
+	rdpr_bytes |= (uint16_t)((uint16_t)~rdpr_bytes) << 8;
+	if( MCF.WriteHalfWord( dev, (intptr_t)&OB->RDPR, rdpr_bytes ) ) goto flashoperr;
 	if( MCF.WaitForFlash(dev) ) goto flashoperr;
 
 	if( MCF.WriteWord( dev, (intptr_t)&FLASH->OBKEYR, FLASH_KEY1 ) ) goto flashoperr;
--- a/minichlink/minichlink.c
+++ b/minichlink/minichlink.c
@@ -2732,6 +2732,7 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 			|| chip == 0x305C0508
 			|| chip == 0x30700508 
 			|| chip == 0x30710508 
+			|| chip == 0x30720508
 			|| chip == 0x30730508
 			|| chip == 0x30740508
 			|| chip == 0x3170B508
@@ -2749,6 +2750,7 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 			|| chip == 0x305C0508
 			|| chip == 0x30700508 
 			|| chip == 0x30710508 
+			|| chip == 0x30720508
 			|| chip == 0x30730508
 			|| chip == 0x30740508
 			|| chip == 0x3170B508
@@ -2766,6 +2768,7 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 			 || chip == 0x305C0508
 			 || chip == 0x30700508 
 			 || chip == 0x30710508 
+			 || chip == 0x30720508
 			 || chip == 0x30730508
 			 || chip == 0x30740508
 			 || chip == 0x3170B508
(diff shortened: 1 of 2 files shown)


minichlink/minichlink.c
```diff
@@ -2700,55 +2703,84 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
...
-			 || chip == 0x30720508
-			 || chip == 0x30740508) {
-				split_code = 0;
+		   if (chip == 0x30300504
+			|| chip == 0x30310504
+			|| chip == 0x305C0508
+			|| chip == 0x30700508 
+			|| chip == 0x30710508 
+			|| chip == 0x30730508
+			|| chip == 0x30740508
```
cnlohr, line 2737: This list is missing 0x30720508. I don't think that's intentional.

minichlink/minichlink.c
```diff
@@ -2700,55 +2703,84 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
...
-			 || chip == 0x30720508
-			 || chip == 0x30740508) {
-				split_code = 1;
+		   if (chip == 0x30300504
+			|| chip == 0x30310504
+			|| chip == 0x305C0508
+			|| chip == 0x30700508 
+			|| chip == 0x30710508 
+			|| chip == 0x30730508
+			|| chip == 0x30740508
```
cnlohr, line 2755: This list is also missing 0x30720508

minichlink/minichlink.c
```diff
@@ -2761,7 +2793,7 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 			 || chip == 0x2081050c
 			 || chip == 0x2082050c
 			 || chip == 0x2083050c) {
-				split_code = 0;
+				split_code = 1;
```
cnlohr, line 2825: Is this really right? Why switch the split codes?

minichlink/minichlink.c
```diff
@@ -2774,7 +2806,7 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 			 || chip == 0x2081050c
 			 || chip == 0x2082050c
 			 || chip == 0x2083050c) {
-				split_code = 1;
+				split_code = 3;
```
cnlohr, line 2838: Also here?

minichlink/minichlink.c
```diff
@@ -2787,7 +2819,7 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 			 || chip == 0x2081050c
 			 || chip == 0x2082050c
 			 || chip == 0x2083050c) {
-				split_code = 2;
+				split_code = 5;
```
cnlohr, line 2851: And here.

minichlink/minichlink.c
```diff
@@ -2812,9 +2844,9 @@ static int DefaultSetSplit(void * dev, enum RAMSplit split) {
 
 	// Option byte b is stored as 16 bits (~b << 8)|b
 	// Mask off upper copy and clear split bits
-	option_bytes &= 0x003F;
-	// Set split code at [7:6]
-	option_bytes |= split_code << 6;
+	option_bytes &= 0x001F;
```
cnlohr, line 2877: @monte-monte do you know anyhting about this one?

monte-monte: @cnlohr no, I've never used split.

DeqingSun: The reason why the split code changed, is that WCH expand ram config bits, from 9:8, to 9:7. Since this expansion happen on the lower side, all previous code need to be stuffed with 1 on the lower side. That is the reason why they all changes.
Also, when we apply splitcode to option byte, we need to change from <<6 to <<5.

minichlink/minichlink.c
DeqingSun: interesting. I added the ID from the reference manual (screenshot in original PR). And it seems ch32v30x_dbgmcu.c was referring to a nonexistent part.

Screenshot 2026-06-04 at 3 14 33 PM

307 never had a TSSOP variation (CH32V307FBP6) on the market.

DeqingSun: 0x30740508 in other split was not added by me, I did not now what it was so I just kept it.

minichlink/minichlink.c
DeqingSun: The ram split was first added at https://github.com/cnlohr/ch32fun/commit/82659c667ca5e8772206fe1545c7f02c5c440ed7
Maybe @iracigt can have a comment?

iracigt: I haven't looked at the updated manual, but this change matches the screenshot above. This is the mask to leave the other bits unchanged. If bit 7 is now part of the field, then yes we should clear it here and make split_code 3 bits instead of 2.

I'd suggest leaving a comment somewhere in this block acknowledging the datasheet changes, just in case a future reader is referencing an older copy.

DeqingSun: comment added

cnlohr:
Mind addressing the other split comments?  And, even though I don't know why the original had the extra packages in the list, mind retaining them, just in case?

DeqingSun:
0x30720508 is added back.
Now there is a comment on why split_code changes and another comment on why the option_bytes also changes

cnlohr:
Thanks! this looks good now.

PR #905, opened 2026-06-04
Fix bug in I2C pin remapping instructions in README

binary1230:
Updated comments for clarity and corrected code for I2C pin remapping.

The original code occasionally failed to remap the 2x I2C pins and which failed i2c communications on ~10% of a batch of 100 boards with CH32V003F4U6 that I just received. both bits have to be set simultaneously, or it seems the remap can fail to take effect.

It must be some kind of manufacturing-batch-specific undefined behavior in certain ch32's, because each chip either consistently worked or consistently failed. it doesn't (it's not like it's flaky after each boot)

I wanted to submit this for initial review. I'm currently verifying this fix on my failed boards, and if everything is good, will give a thumbs up.

I wrote this example originally and got burned in production by my own example code :) What a life

### DIFF
--- a/examples/i2c_slave/README.md
+++ b/examples/i2c_slave/README.md
@@ -34,19 +34,21 @@ Remapping examples:
 // 2 bits [bit 22 and bit 1] of AFIO_PCFR1 register are what control I2C1 pin remapping on ch32v003
 // [high bit, low bit]
 // [I2C1REMAP1, I2C1_RM]
+//
+// Important: always set/clear both bits in a single read-modify-write operation.
+// Two separate writes (one per bit) risk the second read returning stale data on some
+// CH32V003 silicon batches, silently routing I2C to the wrong pins with no other
+// indication of failure.
 
-// [0, 0]				default mapping (SCL on pin PC2, SDA on pin PC1)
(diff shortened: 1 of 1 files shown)


binary1230:
Actually could still be a hardware quirk or, could have been some compiler caching stuff? I may poke around a bit more. This change is still good either way since it's more consistent with how the rest of this repo deals with setting multiple bits at once

binary1230:
sorry for the spam, quick update.

The remap code here might not be the culprit.

I -thought- I had fixed it because my i2c init was working fine when I have the GND+SWIO pin connected to an external esp32 for minichlink (which is how I was testing this fix).  

if I disconnect just the SWIO pin, i2c stops working again.  perhaps it's something to do with the minichlink terminal/printf stuff and exposing some kind of timing bug on my personal ch32 firmware code.

at any rate, the changes here are probably still worth making

cnlohr:
Don't get me wrong, if ordering doesn't matter, then my preference in general would be to do things the way they are on the right, but I don't have a super strong preference either way.

cnlohr:
Yeah, I do like this better.  I'll accept it.

PR #907, opened 2026-06-06
Add vector-free-table example and cleanup ISR setup code.

cnlohr:
1. Add a vector free table example. 
2. Fixup startup assembly.

I tried avoiding t0, to keep the startup registers more consistent, and fix the clobber lists.

This example takes latency from ~558ns in example to 416ns, so, 7 cycles.

NEED TO make sure this doesn't break ch32fun on ch32v305, ch32x035, ch570.


cnlohr:
Tested with rv003usb, and a few other configurations, it seems to be fine.

PR #910, opened 2026-06-11
add status api to ecm example

BogdanTheGeek:
Realised that we mentioned a status api that was actually never ported.
Also investigating some issues when `#define BIG 1` and in general with reliability.


BogdanTheGeek:
got nerdsniped into adding dhcpd so people don't have to set the ip manually

BogdanTheGeek:
@cnlohr I did a quick port from my UIP dhcpd for this, but if you want, I can either build it into sfhip or add it to the repo as a demo?

cnlohr:
@BogdanTheGeek I don't have a strong preference, the DHCP server looks separable enough I don't think I want it in the _main_ `sfhip.h` stack.

PR #912, opened 2026-06-16
Optimize building for ch5xx

monte-monte:
- Remove unneeded package setting for ch5xx
- Fix fallback setting for ch585 clock


cnlohr:
I like this better. Thanks.

PR #913, opened 2026-06-22
Add support for sha1sum back into Makefile for non-Windows os

dwillmore:
In a recent change, the minichlink Makefil changed to useing shasum to generate the file checksums, but Fedora doesn't include that by default--it's unintentionally provided on other Linux OSes through a Perl package.  

Make the script look for shasum or sha1sum on non-Windows platforms and to use that if found.

Changed Windows to use the 'where' command to look for shasum instead of the POSIX 'which' command.


dwillmore:
I was really only worried if the Windows build would pass as that's the least sane environment and I'm not setup to test it.

PR #917, opened 2026-07-01
Updates to USB

monte-monte:
Updated USBFS and USBHS libs to better handle the built-in buffers. Required API changes, updated all USB examples.
Added a universal USB bootloader, compatible with the one in rv003usb. Added functionality to minichlink to support it on ch5xx chips.


monte-monte:
USB FS:
- Now you can add any number of endpoints and it won't waste a space for the buffers in between. (for example: enabled ep1 and ep5 previously would use 5 buffers of space)
- You can enable double buffering, bidirectional, or double buffered bidirectional mode for each endpoint in config. It will use needed number of buffers and set every setting automatically. No examples yet on how to properpy use those modes though.
- Fixed EP > 4 on ch5xx.
- Made usb printf functions weak. This needed fsusb.c to be decoupled from fsusb.h (previosly it was included at the end of the file and built automatically). Now you need to add a fsusb.c to a build list in your Makefile.
- Updated all examples and tested them on each of the supported MCUs (well mostly). Added missing settings to config files.

USB HS:
- Also flexible buffer allocation.
- Added missing double-buffering mode to endpoints (no bidirectional on this peripheral).
- Variable buffer size per each endpoint.

USB bootloader:
- Ported rv003usb bootloader to HW USB.
- Had to write a bunch of binary stubs for minichlink to be able to flash, erase and read ch5xx chips.
- Currently only tested on ch32x035 and ch570. Some extra work needed to make it fully work on ch32v20x and ch32v30x.

minichlink:
- Added a directory for binary stubs that contains .S assembly files and compiled binary in .h files.
- Made a Makefile to automatically build each .S file or separately.
- pgm-b003.c now includes those .h files instead copying their contents directly.
- Fixed bootloader layout definitions for ch32l103 in chips.c
- Fixed some bugs and added needed functions for ch5xx chips in pgm-b003.c

monte-monte:
Will undraft and add minichlink.exe when review will be done.

monte-monte:
Now there is a ``funRebootToBootloader`` function that lets you...reboot into bootloader from your firmware. It works on ch570, ch32x035 and ch32v003, ch32v00x.

monte-monte:
Please review but don't merge before I add minichlink.exe and new version file.

examples_usb/bootloader/bootloader.c
```diff
@@ -0,0 +1,287 @@
...
+__attribute__((section(".runwordpad"))) volatile int32_t runwordpad;
+__attribute__((section(".boot_address"))) volatile uint32_t boot_usercode_address;
+
+uint32_t runwordpadready = 0;
+uint32_t current_scratchpad_size = 128;
+volatile uint32_t cmd_len = 0;
+
+volatile uint8_t reset_timeout = 0;
+
+static inline void asmDelay(int delay) {
```
cnlohr, line 94: Add a comment explaining why this is here instead of using the one built in to ch32fun.

examples_usb/bootloader/bootloader.c
```diff
@@ -0,0 +1,287 @@
...
+	#endif
+	#endif
+
+#elif BOOTLOADER_BTN_PC17 && defined(CH32X03x)
+	GPIOC->CFGXR = (GPIOC->CFGXR & (~(0xf<<(4)))) | ((GPIO_CFGLR_IN_PUPD)<<4);
+	// GPIOC->BSXR = 2<<16; // Setting pull-down this way doesn't work for some reason. But leaving it by default works.
+#endif
+
+#if !defined(DISABLE_BOOTLOAD) || !DISABLE_BOOTLOAD
+	asmDelay(1000000);
```
cnlohr: Add a comment saying how long this is.

examples_usb/bootloader/README.md
```diff
@@ -0,0 +1,5 @@
+# USB Bootloader
```
cnlohr, line 1: Please expand this file greatly, preferably before initial commit.

examples_x035/blink/blink.c
```diff
@@ -2,7 +2,7 @@
 #include <stdio.h>
 
 // use defines to make more meaningful names for our GPIO pins
-#define PIN_1 PA0
+#define PIN_1 PB12
```
cnlohr: y tho

extralibs/hsusb.h
```diff
@@ -413,6 +421,39 @@ typedef enum
...
+		uint8_t * out;
+		uint8_t * rx;
+		uint8_t * in;
+		uint8_t * tx;
+		uint8_t * buffer;
+	};
+	uint16_t size;
+	uint8_t mode;
+	volatile uint8_t busy;
+} USBHS_EP_TypeDef;
```
cnlohr, line 437: If you maintain it I am NOT GOING TO COMPLAIN.  But in general, I prefer the sort of compile-time-only-mechanism for USB control.  But I understand why it's limiting.

cnlohr: Does this mirror something in the hardware?  I don't think it does, but if it does please explain what it mirrors in the hardware.

cnlohr (commented):
In addition to these comments, please thurougly test the legacy USB bootloader with the CH570.

examples_usb/bootloader/bootloader.c
monte-monte: If only I knew, lol.
Ok, I'll measure.

examples_x035/blink/blink.c
monte-monte: Weact board has LED on this GPIO. I asked Willmore if this is ok, and he's chosen which of existing pins to change.

extralibs/hsusb.h
monte-monte: Honestly, I don't understand what you mean in first or the second comment.
Previous way of dealing with buffers was very rigid and not able to accommodate certain common use cases (bi-directional, double-buffering). Since I added changes to fsusb,h, I mirrored them to hsusb.h to have some consistency between the two. USB HS can't do bi-directional endpoints, but it's still beneficial to have a buffer (pointer) named by the function it servers in particular configuration.
Yes, this is a little high-level abstraction-ish, but this is a universal library, I think it can have some. Compile-time-only would be defining buffers manually and passing them to DMA and not having them hidden in an internal struct. But this is not most intuitive way for people to use with examples. I believe our lib is better than EVT example in this regard.

monte-monte:
> please thurougly test the legacy USB bootloader with the CH570.

@cnlohr what do you mean by that?

examples_usb/bootloader/bootloader.c
monte-monte: it saves space.

cnlohr:
> > please thurougly test the legacy USB bootloader with the CH570.
> 
> @cnlohr what do you mean by that?

My normal development process is to:

Send a USB command to my CH570 that reboots it into the factory bootloader.

Flash device.

Command reboot.

Debug.


I'd like to continue to be able to use the factory bootloader for this.

examples_x035/blink/blink.c
cnlohr: Comment it. It's OK to change by default.

extralibs/hsusb.h
cnlohr: My second question is - is the struct you have made something that mirrors any hardware, or is it an abstraction?  If just an abstraction please add a comment that says that this is not mirroring something in the device and is for housekeeping for us to be able to configure and manage a USB system.

monte-monte:
@cnlohr It doesn't impede ISP functionality in any way. You don't have to use this bootloader if you don't want to. But if you do switch to it, you won't feel any difference, while using minichlink.

examples_x035/blink/blink.c
monte-monte: You want me leave a comment why this pin is there, or you want to revert back the change?

extralibs/hsusb.h
monte-monte: This is an abstraction in a way like it's presented. Otherwise the endpoint mode is set in corresponding register and does represent the modes I am using.

cnlohr:
I mean it the other way, I mean, can you validate that the old ISP stuff still works?

cnlohr:
Eh, I'll merge this anyway, if there's any issues we'll fix them in post.

PR #918, opened 2026-07-01
Add support to build on FreeBSD

metanest:
Add riscv64-none-elf and riscv32-unknown-elf in ch32fun.mk.

For minichlink on FreeBSD, use ports-provided hidapi, and system libusb via pkg-config, and include the external hidapi header.

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -92,8 +92,10 @@ void __libc_init_array(void)
 #include <stdint.h>
 #include <ch32fun.h>
 
+#if defined(__FreeBSD__)
 #ifdef putchar
 #undef putchar
+#endif
 #endif
 
 #if defined(CH32H41x)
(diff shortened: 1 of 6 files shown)


dwillmore:
Did not expect my WHICH to get used this fast. :)  Are you sure the naked #ifdef putchar isn't going to break anyone else?  Might want to wrap that in an "if BSD".

dwillmore:
@metanest ping

cnlohr:
@dwillmore up for 👍 this? I'm happy to merge when you're comfortable.

cnlohr:
Err actually looking at this closer, I'm happy with it as is.  Merging.

dwillmore:
Yeah, I was going to ask you. :)  I was happy.

dwillmore:
Glad to see support added for OSs that I don't use.  I hope we get testing from them as well. :)

PR #920, opened 2026-07-01
Add CH32M030 support (framework + blink example)

Enz1ow:
## Summary

Adds first-class support for the **CH32M030** (QingKeV3B, RV32IMC) to the ch32fun
framework. Validated on a **CH32M030G8R7-PMSM** board: the included `blink`
example runs from a cold power-on and toggles PB0 at a steady 2 Hz (scope-measured
2.0033 Hz).

> Flashing support for the M030 lives in a companion minichlink PR (see below) —
> this PR is framework-only and doesn't touch minichlink.

## What's in it

`ch32fun/`:
- `ch32fun.mk` — CH32M030 build dispatch (rv32imc/ilp32, `TARGET_MCU_LD=12`, default package `CH32M030G8R7`).
- `ch32fun.ld` — 64 KB flash / 12 KB RAM memory layout.
- `ch32m030hw.h` — per-chip register map (RM v1.2), following the per-chip-header convention.
- `ch32fun.h` — header selection + HSI/PLL/clock defines.
- `ch32fun.c`:
  - `handle_reset`: M030 uses the V003/X03x/V00x reset variant and sets the QingKeV3B
    core CSRs after vector setup (`corecfgr 0xBC0=0x21`, `intsyscr 0x804=0x03`,
    `inestcr 0xBC1=0x01`, per WCH `startup_ch32m030.S`). Without `corecfgr` the core
    traps in early startup and never reaches `main`.
  - `SystemInit`: dedicated read-modify-write clock bring-up (HSI 8 MHz → fixed-×18
    PLL → HPRE/2 → 72 MHz) that **preserves the factory HSICAL trim** in
    `RCC_CTLR[15:8]`; a wholesale `RCC->CTLR` write wipes it and the PLL never locks.
  - `DelaySysTick`: M030 SysTick is 32-bit compare-mode (`CNT`@0x08 / `CMP`@0x10 are
    32-bit, 0x0C/0x14 reserved); the generic 64-bit `CNT/CMP` access bus-faults. Uses
    the `*L` fields and counts at full HCLK.

`examples_m03x/blink/` — minimal PB0 bring-up demo in a chip-specific example dir
(matching `examples_x00x`, `examples_h41x`, etc.); has `funconfig.h`, so the build CI discovers it.
`README.md` — move ch32m030 to the supported list.

## Testing

- Cold power-on blink at 2 Hz on PB0 (CH32M030G8R7-PMSM), scope-verified (2.0033 Hz).
- Clock state cross-checked against WCH MRS ground truth: `RCC_CTLR=0x03004E83`
  (PLLON+PLLRDY, HSICAL preserved), `RCC_CFGR0=0x1A` (SWS=PLL, HPRE/2), `FLASH_ACTLR=3`.
- `examples/blink` (default ch32v003) still builds — all M030 code is `#if defined(CH32M030)`-guarded.

## Known limitations

- Validated on the **G8R7** package; other M030 packages share the layout but are untested.

## Companion

Flashing the M030 needs the minichlink support in the companion PR: #919 
Companion: #919

Proof_Scope

### DIFF
--- a/ch32fun/ch32fun.mk
+++ b/ch32fun/ch32fun.mk
@@ -130,7 +130,25 @@ else ifeq ($(findstring CH32X03,$(TARGET_MCU)),CH32X03) # CH32X033, X035
 		MCU_PACKAGE:=1
 	endif
 
-	TARGET_MCU_LD:=4		
+	TARGET_MCU_LD:=4
+else ifeq ($(findstring CH32M030,$(TARGET_MCU)),CH32M030) # CH32M030
+	TARGET_MCU_PACKAGE?=CH32M030G8R7
+	CFLAGS_ARCH+=-march=rv32imc \
+		-mabi=ilp32 \
+		-DCH32M030=1
+
+	# MCU Flash/RAM split (M030 always 64K/12K regardless of package)
+	ifeq ($(findstring G8, $(TARGET_MCU_PACKAGE)), G8)
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring F8, $(TARGET_MCU_PACKAGE)), F8)
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring K8, $(TARGET_MCU_PACKAGE)), K8)
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring C8, $(TARGET_MCU_PACKAGE)), C8)
+		MCU_PACKAGE:=1
+	endif
+
+	TARGET_MCU_LD:=12
 else ifeq ($(findstring CH32L103,$(TARGET_MCU)),CH32L103) # CH32L103
 	TARGET_MCU_PACKAGE?=CH32L103C8T6
 	CFLAGS_ARCH+=-march=rv32imac \
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ ch32fun contains:
 
 ## Other family support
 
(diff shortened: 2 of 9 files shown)


ch32fun/ch32fun.mk
```diff
@@ -130,7 +130,25 @@ else ifeq ($(findstring CH32X03,$(TARGET_MCU)),CH32X03) # CH32X033, X035
...
+else ifeq ($(findstring CH32M030,$(TARGET_MCU)),CH32M030) # CH32M030
+	TARGET_MCU_PACKAGE?=CH32M030G8R7
+	CFLAGS_ARCH+=-march=rv32imc \
+		-mabi=ilp32 \
+		-DCH32M030=1
+
+	# MCU Flash/RAM split (M030 always 64K/12K regardless of package)
+	ifeq ($(findstring G8, $(TARGET_MCU_PACKAGE)), G8)
+		MCU_PACKAGE:=1
+	else ifeq ($(findstring F8, $(TARGET_MCU_PACKAGE)), F8)
```
maxgerhardt, line 143: There is no CH32M030F8 chip it all? Huh? Also every possible case here lands in MCU_PACKAGE = 1. All CH32M030 chips have 64K Flash, 12K SRAM. In my opinion, there should the entire ifeq..endif block should be deleted and set the MCU_PACKAGE statically, or don't use it at all for this chip.

From [CH32M030DS0.PDF](https://www.wch-ic.com/downloads/CH32M030DS0_PDF.html)
Image

dwillmore:
Check your whitespace.  We use two character tabs for indenting, not spaces.

For the .mk file, you don't need to parse out the package info if that doesn't change anything in the LD script.  Use the 003 as an example, no package variants in the .mk nor the .ld.  You add a lot of code that does nothing.

In your register defines, you have the enums for GPIO pin configuration which do not reflect what the RM shows.  You need to limit what's in there to what the chip actually does.  Then you need to fix the blink example to use those corrected defines.  

There may be more register define issues.  This really needs to be done by a human by hand one register bit at a time.  It is way too easy to get it wrong otherwise.

dwillmore:
@Enz1ow Are you still working on this?

dwillmore:
This appears to be abandoned by the submitter.

PR #921, opened 2026-07-02
ch5xx.c: pwm: add 16-bit register aliases for CH570/CH572 PWM data registers

eyhc:
**Motivation**

Some PWM channels on the CH570/CH572 support two data register modes: 8-bit and 16-bit, depending on the value of `RB_PWM_CYC_MOD`. The current `PWM_TypeDef` only exposes the registers for 8-bit mode. This PR adds support for the 16-bit mode.

**Changes**

- Replace individual PWMx_DATA 8-bit register definitions with packed unions providing:
  - PWMx_R8_DATA for 8-bit register access.
  - PWMx_R16_DATA for 16-bit register access where supported by the hardware.
- Preserve the original memory layout and register offsets by using packed structs/unions and updating reserved fields accordingly.
- Rename reserved fields (RESERVED0–RESERVED5) to avoid duplicate member names after introducing the unions.

### DIFF
--- a/ch32fun/ch5xxhw.h
+++ b/ch32fun/ch5xxhw.h
@@ -326,20 +326,38 @@ typedef struct {
     __IO uint8_t CONFIG;         // 2H
 #ifdef CH570_CH572
     __IO uint8_t DMA_CTRL;       // 3H
-    __IO uint8_t PWM1_DATA;      // 4H
-    __IO uint8_t PWM2_DATA;      // 5H
-    __IO uint8_t PWM3_DATA;      // 6H
-    uint8_t RESERVED0[5];        // 7H
+    __packed union {
+        __packed struct {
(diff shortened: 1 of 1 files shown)


maxgerhardt:
Looks cool! Do you want to update or create a new PWM example that uses the 16 bit mode, too?

eyhc:
No, if it’s not necessary. This could be the subject of another PR..

cnlohr:
@eyhc I would appreciate that if you could. I spent about 10 hours trying to get PWM working on my CH570Q - a few people noted that it may literally not have the PWM hardware, though.

eyhc:
@cnlohr Sure! but I can’t do it on CH570**Q** because it only supports the PWM0 channel from the Timer and not the `PWMx` peripheral, as shown in the table below:

image

PR #922, opened 2026-07-03
warning: suppress file-level warning when building headers with Zephyr

BOJIT:
Zephyr support for the CH32H417 is being added to zephyr: https://github.com/zephyrproject-rtos/zephyr/pull/111725

A requirement of the zephyr's CI is that warnings are treated as errors.
This PR guards an unconditional warning with the `__ZEPHYR__` macro that all zephyr files are built with.

I'm upstreaming it here as ideally `hal_wch` will exactly follow these upstream files.
Alternatively the entire warning statement could be removed if it's deemed unecessary.

### DIFF
--- a/ch32fun/ch32h41xhw.h
+++ b/ch32fun/ch32h41xhw.h
@@ -13646,7 +13646,11 @@ typedef enum
 #ifndef __ASSEMBLER__
 
 /* Output Maximum frequency selection */
+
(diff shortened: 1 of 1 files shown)


cnlohr:
I'm happy to have vendor-specific opt-in/out's in ch32fun.  Thanks.

PR #923, opened 2026-07-04
Fix SysTick interpretation for H41x

maxgerhardt:
Fixes the bug that the blinky example just crashes after about 5 minutes.

The H41x chip series is a dual-core QingKe V3F and V5F. The V3F starts up first and can release the V5F to run something later.

Technically, each core has its own independent SysTick, but this is not touched here. Both the V3F and V5F have a only 32-bit SYSTICK counter.

grafik
grafik
grafik

Hence they need to be in the codepath for a 32-bit CNT(L).

Without it, and HCLK running at 100 MHz and SYSTICK running at HCKL / 8, a 32-bit overflow occurs after `2^32 / (100_000_000 / 8) seconds = 5 min 44s` which then gets hung up in the delay function.

Runtime discrimination for HartID 0 or 1 and using the core-specific SYSTICK will be an additional improvement, but right now it doesn't break anything.

### DIFF
--- a/ch32fun/ch32fun.c
+++ b/ch32fun/ch32fun.c
@@ -1888,10 +1888,10 @@ void DelaySysTick( uint32_t n )
 #if defined(CH32V003) || defined(CH32V00x)
 	uint32_t targend = SysTick->CNT + n;
 	while( ((int32_t)( SysTick->CNT - targend )) < 0 );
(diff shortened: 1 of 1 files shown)


cnlohr:
Thanks for catching this!

PR #924, opened 2026-07-05
Multiple Makefile fixes for Windows

maxgerhardt:
* Fixes that `rm` is not available on Windows, calls the proper commands instead via `$(RM)` and `$(RMDIR)` . Normalizes some paths for Windows. Fixes `make clean` erroring out on Windows.
* delete duplicate variable creations.
* normalize `CH32FUN` variable to not include final `/`. This was tripping up calls to `del` on Windows.
* Uses `-` in front of the commands to surpress errors.
* Cleans up Minichlink Github action, makes Ubuntu just build the `.exe` target without making `make` think that the host machine is a Windows machine
* adds a Powershell script that generates the `version` file created by the `make hash` target. This seems to not have been updated in quite some time, so the latest version is commited. Does Unix lineending normalization for consistent output.
  * `shasum` does not exist by default on Windows, but `certutil` can be used to generated hashes instead. Here, a powershell script is used with `System.Security.Cryptography.SHA1`. 
* adds libusb dll into the CI ZIP for minichlink, otherwise the binary is not standalone-executable


cnlohr:
Feelsgoodman.  Also, I wonder if we should be internalizing more of the libusb work.  It's frustrating that hidapi is header-only effectively, but libusb is not.

maxgerhardt:
Something I've not yet explored is to just link libusb statically for Windows. That would make it standalone executable.

cnlohr:
@maxgerhardt I've had better luck with just taking source and compiling-in-place for that sort of thing recently.  Libraries are pain

PR #930, opened 2026-07-15
Add LCD example. Bump minichlink

monte-monte:
- Added an LCD example for CH592, CH591F, CH584 and CH585. With a handy .h file that helps using this peripheral.
- Updated minichlink.exe binary and version file (the thing that was not done in previous PR)


cnlohr:
Looks good.

PR #932, opened 2026-07-17
Add chip_id check for b003 programmer

cheyao:
When reading memory address 0x1ffff7c4 on the ch32v203, the chip returns 0xe339e339. As this isn't 0xffffffff, the programmer takes the else branch and mistakes the chip as a v003.

PR #933, opened 2026-07-19
Add write_block stub for v20x and v30x to pgm-b003fun

cheyao:
Added binary stub for the v20x and v30x. It uses a slightly different programming sequence than v003.

Works with a v203:
image
("v20x" print is a debug log that is removed in the pr)

Before the v003 stub doesn't write correctly on the v203, and a 2nd write causes the chip to freeze.

Here's my v203 bootloader if you want to try it out: https://github.com/cheyao/comu/tree/main/bootloader. (It uses the USBD peripheral. Currently should be written to flash address 0, executables should be linked and flashed to flash+0x800)


cheyao:
Oh and this needs to be merged with PR #932 to have the chip family correctly detected. v30x is untested but looking at the EVT the code is exactly the same.

cnlohr:
@monte-monte can you review this?

cnlohr:
Also, @cheyao please remove the .exe from your commit.

monte-monte:
@cnlohr who will update exe file? I asked everyone to do this when committing changes to minichlink

monte-monte:
If cnlohr is strongly against exe in this commit, I still ask you to update ``version`` file with ``make hash`` command, after all changes are made (if any).
Otherwise looks good, I think.

cheyao:
Ok, I've updated the `version` file and removed the change in `minichlink.exe` too

cnlohr:
@monte-monte I would only be comfortable if someone on the core team builds the .exe, otherwise we open the door for possible worms and viruses.

monte-monte:
Okay, I understand. Maybe it would be better if github could build exe after commit is merged. If this even possible.

cnlohr:
Is everything else about this PR OK, @monte-monte ?

monte-monte:
@cnlohr my questions left unanswered, but they are not critical, so I think it's all good.

PR #934, opened 2026-07-22
Fix linke not working in windows

monte-monte:
Fixes issue with minichlink not being able to open linke device in windows.


DeqingSun:
Thank you, ```libusb_free_device_list( list, 1 );``` was added by me .Removing it does not affect the functionaity in Mac.

PR #935, opened 2026-07-24
Update USBD library to match USBFS API & support HID

cheyao:
Tested with examples:
- bootloader
- usbfs_bulk_echo
- usbfs_cdc_tty
- usbfs_device
- usbfs_cdc_ecm (Not 100% sure if I'm accessing the network interface correctly, not a networking guy)

Update:
- usbfs_msc now works

Limitations:
- USBD peripheral is fundamentally different from USBFS, so anything that access raw USBFS will take effort to port (usbfs_cdc_uart)

More info on discord

PS. The code style is base on clang-format in this repo. (My usual style isn't this xD)

### DIFF
--- a/extralibs/usbd.c
+++ b/extralibs/usbd.c
@@ -1,534 +1,796 @@
 #include "usbd.h"
-#include <string.h>
+#include "ch32fun.h"
+#include "usb_config.h"
 
-// loosely based on https://github.com/ArcaneNibble/wch-uf2/blob/main/bootloader.c
-// this would not exist if it was not for ArcaneNibble's efforts (I don't know how else to credit that)
+#include <stdio.h>
+
+// clang-format off
+#define ANYPRINTF	(defined( FUNCONF_USE_DEBUGPRINTF ) && FUNCONF_USE_DEBUGPRINTF) || \
+					(defined( FUNCONF_USE_UARTPRINTF ) && FUNCONF_USE_UARTPRINTF) || \
+					(defined( FUNCONF_USE_USBPRINTF ) && FUNCONF_USE_USBPRINTF)
+
+volatile uint8_t usb_debug = 0;
+// Uncomment to enable debugging
+// #define DEBUG
+#define USBD_DEBUGPRINTF( ... ) {if ( usb_debug ) { printf( __VA_ARGS__ ); }}
+#define min(a, b) (((a)<(b))?(a):(b))
+// clang-format on
 
 struct _USBState USBDCTX;
 
-#define BTABLE_OFFSET 0
+void USBD_InternalFinishSetup();
 
-static void PMA_Write(uint16_t pma_addr, const uint8_t *buffer, int len) {
-	volatile uint32_t *pma = (volatile uint32_t *)(USBD_PMA_BASE + pma_addr);
-	for (int i = 0; i < len; i += 2) {
-		uint16_t val = buffer[i];
-		if (i + 1 < len) {
-			val |= (uint16_t)buffer[i + 1] << 8;
-		}
-		*pma++ = val;
-	}
-}
+static inline void SetEPR_Status( const int ep, const uint16_t mask, const uint16_t value )
+{
+	const uint16_t reg = USBD->EPR[ep];
+	const uint16_t current_stat = reg & mask;
+	// Which bits we need to toggle
+	const uint16_t toggle = current_stat ^ value;
 
-static void PMA_Read(uint16_t pma_addr, uint8_t *buffer, int len) {
-	volatile uint32_t *pma = (volatile uint32_t *)(USBD_PMA_BASE + pma_addr);
-	for (int i = 0; i < len; i += 2) {
-		uint32_t val = *pma++;
-		buffer[i] = val & 0xFF;
-		if (i + 1 < len) {
-			buffer[i + 1] = (val >> 8) & 0xFF;
-		}
-	}
-}
-
-static void SetPMA_TxCount(int ep, int count) {
-	int pma_idx = (ep *4) + 1;
-	((uint32_t*)USBD_PMA_BASE)[pma_idx] = count;
-}
-
-static int GetPMA_RxCount(int ep) {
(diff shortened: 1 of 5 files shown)


cheyao:
Also fixed the usbd demo

cnlohr:
@biemster @monte-monte can you review this?

biemster:
i can review when I'm back from holiday.
One thing i can say now is that msc should work, i use that with the current usbd extralib. but that doesn't have to be a blocker for this pr although it is a regression.

cnlohr:
I will leave this open for you.

cheyao:
@cnlohr I've just updated the lib to fix a few bugs. Now all the examples that don't heavily rely on USBFS work!

extralibs/usbd.h
```diff
@@ -1,178 +1,183 @@
-#ifndef _USBD_H
-#define _USBD_H
+#ifndef BOOTLOADER_H
+#define BOOTLOADER_H
```
maxgerhardt: Why is the `usbd.h` file using `BOOTLOADER_H` as its include guard, I would have left the original `_USBD_H`.

maxgerhardt (changes requested):

extralibs/usbd.h
cheyao: My bad, copied over code from my usbd bootloader so I guess it carried over. Should be fixed

cnlohr:
Just need a 👍 from @monte-monte and @biemster

extralibs/usbd.c
```diff
@@ -1,544 +1,857 @@
...
-}
+int USBDSetup( void )
+{
+	// Initialize USB pins to pull-down to avoid mis-detection
+	GPIOA->CFGHR &= ~( ( 0xf << ( 4 * 3 ) ) | ( 0xf << ( 4 * 4 ) ) );
+	GPIOA->CFGHR |= ( GPIO_Speed_2MHz | GPIO_CNF_OUT_PP ) << ( 4 * 3 ) | ( GPIO_Speed_2MHz | GPIO_CNF_OUT_PP )
+	                                                                         << ( 4 * 4 );
+	GPIOA->BSHR = ( 1 << ( 16 + 11 ) ) | ( 1 << ( 16 + 12 ) );
+
+	// We need 42MHz clock for USB
```
biemster: 48 I think? Just a nit but it might confuse people reading this

extralibs/usbd.h
biemster: Is this resolved and the github diff viewer is tardy for me?

extralibs/usbd.h
```diff
@@ -1,178 +1,182 @@
...
+
+#define USBD_EP_MODE_TX 4
+#define USBD_EP_MODE_RX 8
+#define USBD_EP_MODE_BDIR 12
+
+// Magic value for ack
+#define USBD_SEND_ACK ( 0xFFFF )
+
+#define USBFS_EP_MODE_TX "Are you sure you want to use USB FS EPs?"
+#define USBFS_EP_MODE_RX "Are you sure you want to use USB FS EPs?"
```
biemster: How does this work? Maybe rewrite this from a question into a suggestion which macro to use instead.

biemster:
From the looks of it this is great. It's basically a complete replacement of the USBD extralib, which is probably a good idea in the first place with the new explicit code generation guidelines, the original usbd extralib was largely AI generated.
If it doesn't break any demo's and works properly I definitely give a :+1: 
One thing to note though, the USBFS extralib is compiled by just including `fsusb.h` because on the bottom it has `#include fsusb.c`. This was explicitly done in the usbd extralib too, but this PR will remove that so `ADDITIONAL_C_FILES` is now needed in the Makefile. I rather don't have that, so I'd like @cnlohr and @monte-monte opinion on that as well.

extralibs/usbd.c
cheyao: Opps, my bad. Should be fixed (when I force push)

extralibs/usbd.h
cheyao: When porting USBFS code to the USBD lib, I've forgot to change USBFS_EP_MODE_TX to USBD_EP_MODE_TX several times and the code complied without errors, so I decided to add these defines. Though on a second thought I just reworded the error, changed it to a `#warning` and also made defining `USBD_SILENCE_EP_MODE` silence the warning.

cheyao:
> One thing to note though, the USBFS extralib is compiled by just including `fsusb.h` because on the bottom it has `#include fsusb.c`. This was explicitly done in the usbd extralib too, but this PR will remove that so `ADDITIONAL_C_FILES` is now needed in the Makefile. I rather don't have that, so I'd like @cnlohr and @monte-monte opinion on that as well.

Wait is this still true? I just looked again in the `fsusb.h` lib and there seems to be no additional includes, all makefiles in the examples folder also uses `ADDITIONAL_C_FILES`. (Maybe the behavior changed?)

monte-monte:
no, fsusb.h is no longer includes .c file. I changed it because otherwise I couldn't use "weak" attribute for the functions. And also it seems more correct way to do it anyway.

biemster:
> > One thing to note though, the USBFS extralib is compiled by just including `fsusb.h` because on the bottom it has `#include fsusb.c`. This was explicitly done in the usbd extralib too, but this PR will remove that so `ADDITIONAL_C_FILES` is now needed in the Makefile. I rather don't have that, so I'd like @cnlohr and @monte-monte opinion on that as well.
> 
> Wait is this still true? I just looked again in the `fsusb.h` lib and there seems to be no additional includes, all makefiles in the examples folder also uses `ADDITIONAL_C_FILES`. (Maybe the behavior changed?)

Indeed you are right, this changed a while ago without me noticing. In that case, as far as I'm concerned, this PR is good to go! Thanks a lot for this work, looks great!

cnlohr:
@cheyao you still have a requested change by @maxgerhardt

cheyao:
Opps, somehow my changes didn't update. All should be good now @cnlohr

PR #937, opened 2026-07-28
add encoder example for x035

NyxCode:
This PR adds an example for using a rotary encoder read by TIM1 or TIM2 on the x035. Like with the PD example, the functionality is in a generic `rotary_encoder.h` that can be copied into other projects. 
This has not been fully tested yet, since i'm not working with an eval board with those pins exposed. If that's a problem, we can wait until I get one or someone else tests this.

### DIFF
--- a/examples_x035/encoder/Makefile
+++ b/examples_x035/encoder/Makefile
@@ -1,0 +1,9 @@
+all : flash
+
+TARGET:=encoder
+TARGET_MCU:=CH32X035
+
+include ../../ch32fun/ch32fun.mk
+
+flash : cv_flash
+clean : cv_clean
--- a/examples_x035/encoder/encoder.c
+++ b/examples_x035/encoder/encoder.c
@@ -1,0 +1,21 @@
+#include "ch32fun.h"
+#include "rotary_encoder.h"
+#include <stdio.h>
+
+int main()
+{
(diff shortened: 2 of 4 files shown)


NyxCode:
While toying with this, i noticed that some pins are not currently accessible on the X035, e.g. PB21, PB16, PB17, and PC19.

cnlohr:
@NyxCode this is shaped much more like the EVT code than I think I'm comfortable with.  The examples should be a plain direction of how to use the peripheral, not have configuration structures.

cnlohr:
(This isn't a NO, but, it is not particularly the direction we want to take things with ch32fun)

NyxCode:
Alright. This is, like the USBPD example, an abstraction over the raw hardware. If that's out-of-scope for the repo, feel free to close the PR.

cnlohr:
USBPD is at the extreme end of what I am willing to accept.  We should probably have expressed that somewhere.

cnlohr:
What I mean is it would drag us _further_ away from where we want to go.

cnlohr:
@NyxCode are you willing to roll back the way you did this and do it in a more ch32fun centric way (#define's to reconfigure, etc.), I am looking at updating the USBPD stuff so people don't get confused like you in the future.

PR #941, opened 2026-08-05
Update CH570/CH572 CMP_TypeDef register mapping

eyhc:
This PR updates the CH570/CH572 comparator register structure to expose both byte-wise and 32-bit access to the `CTRL` register.

The structure now provides `CTRL0` to `CTRL3` fields matching the register definitions (`R8_CMP_CTRL_0` to `R8_CMP_CTRL_3`) while keeping the existing `CTRL` 32-bit access for compatibility.


cnlohr:
That seems fine.

PR #943, opened 2026-08-17
V203 - Fixed invalid CAN Baud rates

DevinJM3:
I was playing around with the can example code and it seemed to work fine. Upon attaching my logic analyzer I noticed that the BAUD_1Mbps setting actually only worked at 750kbps. Upon digging into the datasheet I noticed that 2 different CAN baud rates are specified.

## Baud Rate Calculations in the Datasheet
---
image

---

image

---

Upon playing around with the values and testing at 1Mbps, i can verify that the bottom equation is the correct equation. I have been able to confirm that it is indeed running at 1Mbps on my Analog Discovery 3 Logic Analyzer.

I have only verifed that the BAUD_1Mbps is the correct speed at 144Mhz on the HSE.
image

### DIFF
--- a/examples_v20x/canbus_network/canbus_network.c
+++ b/examples_v20x/canbus_network/canbus_network.c
@@ -8,6 +8,7 @@
 // These values are calculated based on the system clock clock being 144MHz
 // The print function bellow will print the actual baud rate
 // if the clock or divider is different.
+#if FUNCONF_SYSTEM_CORE_CLOCK == 144000000
 //                       TS1         TS2            BRP
 #define BAUD_25kbps  ((5 << 16) | (4 << 20) | (479 / AHB1_DIV))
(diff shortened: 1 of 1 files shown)


BogdanTheGeek:
I think this might have been due to my clock being 120mhz when I was setting this up, because it definitely worked with an esp32 set to the same baud.
I think the ideal fix would calculate it from pclk like we do for the clock divider.
Someone needs to test it with multiple configs

cnlohr:
I agree, @DevinJM3 can you please `#if FUNCONF_SYSTEM_CORE_CLOCK == your cpu freq` around this?  Or use that to auto calc?

DevinJM3:
@cnlohr I added the #ifdef for you

cnlohr:
Looks great.

PR #944, opened 2026-08-19
Guard more statement-like macros with a do{}while(0)

simon-budig:
Some of the macros are statements and look like function calls. However, in certain circumstances this creates unexpected behaviour:

```
if (some_condition)
  funPinMode(PA2,GPIO_CFGLR_IN_PUPD);
else
  funPinMode(PA2,GPIO_CFGLR_OUT_2Mhz_PP);
```

this fails to compile since it gets expanded with an semicolon in front of the else, basically putting an empty statement after the first conditional block, terminating the if statement too early.

The patch remedies this with the well established do{}while(0) pattern.

### DIFF
--- a/ch32fun/ch32fun.h
+++ b/ch32fun/ch32fun.h
@@ -968,14 +968,14 @@ RV_STATIC_INLINE void funPinMode(u32 pin, GPIOModeTypeDef mode)
 #define funDigitalWrite( pin, value ) do{ GpioOf( pin )->BSHR = 1<<((!(value))*16 + ((pin) & 0xf)); }while(0)
 
 #if defined(CH32X03x)
-#define funGpioInitAll() { RCC->APB2PCENR |= ( RCC_APB2Periph_AFIO | RCC_APB2Periph_GPIOA | RCC_APB2Periph_GPIOB | RCC_APB2Periph_GPIOC ); }
-#define funPinMode( pin, mode ) { *((&GpioOf(pin)->CFGLR)+((pin&0x8)>>3)) = ( (*((&GpioOf(pin)->CFGLR)+((pin&0x8)>>3))) & (~(0xf<<(4*((pin)&0x7))))) | ((mode)<<(4*((pin)&0x7))); }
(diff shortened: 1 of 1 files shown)


sellicott:
Would it be prudent to put the `pin` parameter into a temporary to prevent multiple values getting assigned to it, or multiple function calls?

Scenario 
```C
 funPinMode(get_pin(/*...*/),GPIO_CFGLR_IN_PUPD);
```
Currently `get_pin` would get called multiple times.

cnlohr:
Can you provide an example where that could happen?

sellicott:
That was the scenario in the comment. Is it likely to happen, probably not.

cnlohr:
OH! I see what you meant.  Like if `pin` was a volatile.

cnlohr:
Eh, it's ok.  Let's just roll with this.

simon-budig:
> Would it be prudent to put the `pin` parameter into a temporary to prevent multiple values getting assigned to it, or multiple function calls?
> 
> Scenario
> 
> ```c
>  funPinMode(get_pin(/*...*/),GPIO_CFGLR_IN_PUPD);
> ```
> 
> Currently `get_pin` would get called multiple times.

Macros are tricky and the macros here also have more issues not addressed with my pull request.

Usually macro arguments should always be used with parentheses, especially in conjunction with operators. Right now this would go wrong:

```
int base_pin = 7;
funPinMode (base_pin, GPIO_CFGLR_IN_PUPD);
funPinMode (base_pin+1, GPIO_CFGLR_IN_PUPD);
```

(the example is somewhat contrived, but you get the idea)

In the macro the "base_pin+1" gets expanded as (base_pin+1&0x08) and all hell breaks loose due to operator precedence.  (EDIT: I got that wrong. It would work since + has *higher* precedence than &. In other scenarios it'd go wrong:
`  funPinMode (hw_variant == 1 ? PA2 : PA3, GPIO_CFGLR_IN_PUPD);
`)

The problem you mentioned is multiple evaluation of the argument. This is also fun with operators with side effects:

```
int pin = 7;
funPinMode (pin++, GPIO_CFGLR_IN_PUPD);
funPinMode (pin++, GPIO_CFGLR_IN_PUPD);
```

(note that I do not advocate doing code like that, but sometimes stuff like this happens and then search for the error is nasty...)

I might have a look at these in the next few days. However, at some point it might be better to replace the macros with real functions.

PR #946, opened 2026-08-25
I am pretty sure this is a typo.  Any chance @montemonte could review?

cnlohr:
@monte-monte or @biemster whichever has a system handy they could test the functioning of the CDC UART demo, that would be really helpful.


monte-monte (approved):
Yes, thank you for fixing it.

monte-monte (commented):
this is odd. They set this bit for every USB chip in EVT, for ch570 too.

PR #954, opened 2026-09-09
Change GPIO pin definitions on ch5xx

monte-monte:
**BREAKING CHANGES**
This makes GPIO pins on ch5xx work the same way as on ch32. Before they were defined as masks, now you can iterate through them while usin ``funDigitalWrite()`` and other macros.
I've looked through all examples and found only my encoder example to need changes.
Please tell me if this is good enough.


ch32fun/ch32fun.h
```diff
@@ -916,13 +916,13 @@ extern "C" {
...
+#define GPIO_SetBits(pin)             (*(&R32_PA_SET + OFFSET_FOR_GPIOB(pin)) =  (1<<(pin & ~PB)))
 #endif
-#define GPIO_ResetBits(pin)           (*(&R32_PA_CLR + OFFSET_FOR_GPIOB(pin)) =  (pin & ~PB))
-#define GPIO_InverseBits(pin)         (*(&R32_PA_OUT + OFFSET_FOR_GPIOB(pin)) ^= (pin & ~PB))
-#define GPIO_ReadPortPin(pin)         (*(&R32_PA_PIN + OFFSET_FOR_GPIOB(pin)) &  (pin & ~PB))
+#define GPIO_ResetBits(pin)           (*(&R32_PA_CLR + OFFSET_FOR_GPIOB(pin)) =  (1<<(pin & ~PB)))
+#define GPIO_InverseBits(pin)         (*(&R32_PA_OUT + OFFSET_FOR_GPIOB(pin)) ^= (1<<(pin & ~PB)))
+#define GPIO_ReadPortPin(pin)         (*(&R32_PA_PIN + OFFSET_FOR_GPIOB(pin)) &  (1<<(pin & ~PB)))
 #define funDigitalRead(pin)           !!GPIO_ReadPortPin(pin)
 #define funDigitalWrite( pin, value ) do{ if((value)==FUN_HIGH){GPIO_SetBits(pin);} else if((value)==FUN_LOW){GPIO_ResetBits(pin);} }while(0)
```
lhartmann: Direct compare `(value)==FUN_HIGH` will fail on any value other than 1. In C zero is false, anything else is true, and DigitalWrite usually honors that. We use the same logic for other parts, by testing `!(value)`, in this code:

https://github.com/cnlohr/ch32fun/blob/50b6e591f466231bf3e3ea2c184b3c686e93d755/ch32fun/ch32fun.h#L968

Maybe the logic below (expanded for readability):
```
#define funDigitalWrite( pin, value ) do{ \
    if((value)==FUN_LOW) { \
        GPIO_ResetBits(pin); \
    } else { \
        GPIO_SetBits(pin); \
    } \
}while(0)
```

Or simply replace by `(value)==FUN_HIGH` by `(value)!=FUN_LOW` and achieve the same effect.

